PR #23, opened 2020-05-01
Optional Api Url Parameter

jbraiuka:
Added in optional api url parameter

### DIFF
--- a/test/test_utils.go
+++ b/test/test_utils.go
@@ -24,7 +24,7 @@ func AccTestEnvVarsCheck(t *testing.T) {
 }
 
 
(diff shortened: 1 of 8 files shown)


test/test_utils.go
```diff
@@ -22,3 +22,12 @@ func AccTestEnvVarsCheck(t *testing.T) {
 		t.Fatal("IC_PROV_VPC_ID for provisioning API must be set for acceptance tests")
 	}
 }
+
+
+func getoptionalenv(key, fallback string) string {
```
arjunrajshekhar, line 27: Could this be camel cased?

PR #40, opened 2020-08-04
Changed kafka_user_list to be of data source type

h3nd24:
Changing the type of kafka_user_list from resource to data source

### DIFF
--- a/instaclustr/resource_kafka_user.go
+++ b/instaclustr/resource_kafka_user.go
@@ -68,6 +68,7 @@ func resourceKafkaUserCreate(d *schema.ResourceData, meta interface{}) error {
 			// user is already set, so we don't change anything
 			d.SetId(fmt.Sprintf("%s-%s", cluster_id, username))
 			log.Printf("[INFO] Kafka user %d already exists in %s.", username, cluster_id)
+			return nil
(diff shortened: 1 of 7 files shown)


instaclustr/resource_kafka_user.go
```diff
@@ -83,7 +58,7 @@ func resourceKafkaUserCreate(d *schema.ResourceData, meta interface{}) error {
 		return fmt.Errorf("[Error] Cluster %s is not RUNNING.", cluster_id)
 	}
 		
-	// just use linear search for now
+	// just use linear search for now to check if the user going to be created is already in the system
```
arjunrajshekhar, line 61: For the block -
```
for _, str := range usernameList {
		if str == username {
			// user is already set, so we don't change anything
			d.SetId(fmt.Sprintf("%s-%s", cluster_id, username))
			log.Printf("[INFO] Kafka user %d already exists in %s.", username, cluster_id)
		}
	}
```
Could this exit the method early? Otherwise it will continue with the create kafka user request.

arjunrajshekhar:
An update to the readme would be required to reflect that `instaclustr_kafka_user_list` is a data source instead of a resource

PR #43, opened 2020-08-21
Security group firewall rules

jbraiuka:
Pull request to allow provisioning firewall rules with Security Groups.

### DIFF
--- a/README.md
+++ b/README.md
@@ -164,12 +164,14 @@ truststore|Base64 encoded version of the TLS trust store (in JKS format) used to
 
 ### Resource:  `instaclustr_firewall_rule`                             
 A resource for managing cluster firewall rules on Instaclustr Managed Platform. A firewall rule allows access to your Instaclustr cluster.
+Note: Either rule_cidr OR rule_security_group_id must be provided per rule (but not both)
 
 #### Properties
 Property | Description | Default
 ---------|-------------|--------
 cluster_id|The ID of an existing Instaclustr managed cluster|Required
-rule_cidr|The network to add to your cluster firewall rule. Must be a valid IPv4 CIDR|Required
+rule_cidr|The network to add to your cluster firewall rule. Must be a valid IPv4 CIDR|Optional
+rule_security_group_id|The Peered AWS VPC Security Group id to your cluster firewall rule|Optional
 rules|List of rule types that the specified network is allowed access to. See below for rule options.|Required
 
 `rules`
--- a/README.md
+++ b/README.md
@@ -164,14 +164,14 @@ truststore|Base64 encoded version of the TLS trust store (in JKS format) used to
 
 ### Resource:  `instaclustr_firewall_rule`                             
 A resource for managing cluster firewall rules on Instaclustr Managed Platform. A firewall rule allows access to your Instaclustr cluster.
-Note: Either rule_cidr OR rule_security_group_id must be provided per rule (but not both)
+Note: Either `rule_cidr` OR `rule_security_group_id` must be provided per rule (but not both)
 
 #### Properties
 Property | Description | Default
 ---------|-------------|--------
 cluster_id|The ID of an existing Instaclustr managed cluster|Required
-rule_cidr|The network to add to your cluster firewall rule. Must be a valid IPv4 CIDR|Optional
-rule_security_group_id|The Peered AWS VPC Security Group id to your cluster firewall rule|Optional
+rule_cidr|The network to add to your cluster firewall rule. Must be a valid IPv4 CIDR (e.g. 123.4.5.67/32) |Optional
+rule_security_group_id|The Peered AWS VPC Security Group id to your cluster firewall rule (e.g. sg-12345678) |Optional
 rules|List of rule types that the specified network is allowed access to. See below for rule options.|Required
 
 `rules`
--- a/README.md
+++ b/README.md
@@ -298,7 +298,7 @@ A schema change has been made from 0.0.1 which no longer supports the `bundles` 
 Firstly thanks!  We value your time and will do our best to review the PR as soon as possible. 
 
 1. [Install golang](https://golang.org/doc/install#install)
-2. Clone repository to: $GOPATH/src/github.com/instaclustr/terraform-provider-instaclust
+2. Clone repository to: $GOPATH/src/github.com/instaclustr/terraform-provider-instaclustr
 3. Build the provider by `$ make build`
 4. Run the tests by `$ make test`
 5. Set up all of the environmental variables listed [below](#acceptance-test-environment-variables) to prepare for acceptance testing.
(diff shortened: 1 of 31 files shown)


instaclustr/resource_firewall_rule.go
```diff
@@ -53,11 +57,23 @@ func resourceFirewallRuleCreate(d *schema.ResourceData, meta interface{}) error
 
 		rules = append(rules, RuleType{Type: aRule})
 	}
-
-	rule := FirewallRule{
-		Network: d.Get("rule_cidr").(string),
-		Rules:   rules,
+	var rule FirewallRule
```
arjunrajshekhar, line 60: [Desired, simplicity] - using return early return often + merging 2 blocks, could this logic be simplified to 

```
if d.Get("rule_cidr") != "" && d.Get("rule_security_group_id") != "" {
		return fmt.Errorf("[Error] Error creating firewall rule: Only one of Security Group of Rule Cidr can be provided per rule")
	}

	if d.Get("rule_cidr") == "" && d.Get("rule_security_group_id") == "" {
		return fmt.Errorf("[Error] Error creating firewall rule: either one of Security Group of Rule Cidr is required")
	}

	var rule = FirewallRule{
		Network:         d.Get("rule_cidr").(string),
		SecurityGroupId: d.Get("rule_security_group_id").(string),
		Rules:           rules,
	}
````

File: instaclustr/resource_firewall_rule.go [hunk 2]
```diff
@@ -115,10 +132,17 @@ func resourceFirewallRuleDelete(d *schema.ResourceData, meta interface{}) error
 
 		rules = append(rules, RuleType{Type: aRule})
 	}
-
-	rule := FirewallRule{
+	var rule FirewallRule
```
arjunrajshekhar, line 135: [Desired, simplicity] - Could this be simplified to -

```
	var rule = FirewallRule{
		Network: d.Get("rule_cidr").(string),
		SecurityGroupId: d.Get("rule_security_group_id").(string),
		Rules:   rules,
	}
```

test/resource_pci_cluster_test.go
```diff
@@ -51,13 +51,14 @@ func TestAccPCIClusterResize(t *testing.T) {
 	validConfig, _ := ioutil.ReadFile("data/valid_with_resizable_pci_cluster.tf")
 	username := os.Getenv("IC_USERNAME")
 	apiKey := os.Getenv("IC_API_KEY")
-	oriConfig := fmt.Sprintf(string(validConfig), username, apiKey)
+	hostname := getOptionalEnv("IC_API_URL", instaclustr.DefaultApiHostname)
+	oriConfig := fmt.Sprintf(string(validConfig), username, apiKey, hostname)
```
arjunrajshekhar, line 55: [Question] - what does `oriConfig` mean/stand for?

README.md
```diff
@@ -164,12 +164,14 @@ truststore|Base64 encoded version of the TLS trust store (in JKS format) used to
 
 ### Resource:  `instaclustr_firewall_rule`                             
 A resource for managing cluster firewall rules on Instaclustr Managed Platform. A firewall rule allows access to your Instaclustr cluster.
+Note: Either rule_cidr OR rule_security_group_id must be provided per rule (but not both)
```
peter-toms, line 167: [required] could we add an example somewhere showing what an example `rule_security_group_id` would look like? (possibly an indication of where to find it easily in aws as well)

README.md
```diff
@@ -298,7 +300,7 @@ A schema change has been made from 0.0.1 which no longer supports the `bundles`
 Firstly thanks!  We value your time and will do our best to review the PR as soon as possible. 
 
 1. [Install golang](https://golang.org/doc/install#install)
-2. Clone repository to: $GOPATH/src/github.com/instaclustr/terraform-provider-instaclust
+2. Clone repository to: $GOPATH/src/github.com/instaclustr/terraform-provider-instaclustr
```
peter-toms, line 303: oh my ... good catch

instaclustr/resource_firewall_rule.go
```diff
@@ -23,7 +23,11 @@ func resourceFirewallRule() *schema.Resource {
 
 			"rule_cidr": {
```
peter-toms, line 24: [desired] this could def cause some confusion if both marked as optional ? ie it could pass the terraform validation yet fail at the API end.  I see we have tried todo that logic within the `resourceFirewallRuleCreate` call.  Would it be cleaner todo it within a custom `ValidateFunc` ? and thus catch it before proceeding into the api calling code -> https://github.com/hashicorp/terraform-plugin-sdk/blob/9f0df37a8fdb2627ae32db6ceaf7f036d89b6768/helper/schema/schema.go#L310-L312

peter-toms (changes requested):

README.md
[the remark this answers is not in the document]
jbraiuka: Oh yep that makes sense - will action

instaclustr/resource_firewall_rule.go
jbraiuka: I will check if that works - if it does then i will do this :)

instaclustr/resource_firewall_rule.go
jbraiuka: If it still omits the parameter from the json request if its set to "" then i should be able to simplify.

test/resource_pci_cluster_test.go
jbraiuka: Truthfully i don't know - i would assume original config but i will clarify and update

README.md
jbraiuka: Sure - I can add something like "e.g. sg-12345678" ? 
Not sure if we want to go down the rabbithole of explaining where to find it, at this point we are assuming you have already vpc peered and setup your route tables accordingly - it might be an invalid assumption but i would assume if you have come that far you know how to find a security group? 
We might have some documentation i can link to?

README.md
jbraiuka: #OpenSourceContributer is going straight to the linked in for this one 😄

instaclustr/resource_firewall_rule.go
jbraiuka: @arjunrajshekhar  did raise that, but then pointed out you can only validate an individual field, not validate across multiple fields? 
Happy to refactor if thats incorrect.

test/resource_pci_cluster_test.go
arjunrajshekhar: Haha yeah I think that makes sense, also just realized you didn't introduce this variable, sorry for the comment! 😅

arjunrajshekhar (approved):
Thanks Jordan, looks good!

peter-toms (approved):

PR #53, opened 2020-09-28
Add provider variable to kms encryption keys, fix cluster tags

jbraiuka:
This pr provides a provider variable to the Encryption keys object. 

We also fixed an issue with tags having an incompatible type to the Instaclustr API.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,5 +1,5 @@
 BIN_NAME="terraform-provider-instaclustr"
-VERSION=v1.5.1
+VERSION=v1.6.0
 
 .PHONY: install clean all build test testacc
 
--- a/Makefile
+++ b/Makefile
@@ -1,5 +1,5 @@
 BIN_NAME="terraform-provider-instaclustr"
-VERSION=v1.5.0
+VERSION=v1.5.1
 
 .PHONY: install clean all build test testacc
 
--- a/README.md
+++ b/README.md
@@ -238,17 +238,8 @@ key_provider|For customers running in their own account. Value specifying the th
 #### Example
 ```
 resource "instaclustr_encryption_key" "example_encryption_key" {
-    alias: "virginia 1"
-    arn: "arn:aws:kms:us-east-1:123456789012:key12345678-1234-1234-1234-123456789abc"
-}
-```
-
-Example of an additional provider:
-```
-resource "instaclustr_encryption_key" "example_encryption_key" {
-    alias: "virginia 1"
-    arn: "arn:aws:kms:us-east-1:123456789012:key12345678-1234-1234-1234-123456789abc"
-    key_provider: "INSTACLUSTR"
+    alias = "virginia 1"
+    arn = "arn:aws:kms:us-east-1:123456789012:key12345678-1234-1234-1234-123456789abc"
 }
 ```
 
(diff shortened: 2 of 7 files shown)


README.md
```diff
@@ -233,6 +233,7 @@ Property | Description | Default
 key_id|Internal ID of the KMS encryption key. Can be found via GET to `https://api.instaclustr.com/provisioning/v1/encryption-keys`|""
 alias|KMS key alias, a human-readibly identifier specified alongside your KMS ARN|""
 arn|KMS ARN, identifier specifying provider, location and key in a ':' value seperated string|""
+provider|value specifying the the provider account, same as the cluster_provider|INSTACLUSTR
```
jbraiuka, line 236: I don't like the wording here. anyone else got any ideas?

Makefile
```diff
@@ -1,5 +1,5 @@
 BIN_NAME="terraform-provider-instaclustr"
-VERSION=v1.5.0
+VERSION=v1.5.1
```
arjunrajshekhar, line 2: [Required, correctness] - since this isn't a bugfix this should be a minor version bump

README.md
```diff
@@ -233,6 +233,7 @@ Property | Description | Default
 key_id|Internal ID of the KMS encryption key. Can be found via GET to `https://api.instaclustr.com/provisioning/v1/encryption-keys`|""
 alias|KMS key alias, a human-readibly identifier specified alongside your KMS ARN|""
 arn|KMS ARN, identifier specifying provider, location and key in a ':' value seperated string|""
+key_provider|value specifying the the provider account, same as the cluster_provider|INSTACLUSTR
```
arjunrajshekhar, line 236: [Desired] - Minor typo here, should capitalize "value"
Also should "cluster_provider" should be replaced with `instaclustr_cluster.cluster_provider.account_name`?

h3nd24, line 236: The wording is a bit strange to me, if it's the same as the cluster_provider then why do we need key_provider?

h3nd24:
Based on my past PRs to TF, I would say this still needs example and test.

arjunrajshekhar (changes requested):

README.md
jordanbraiuka: Yea - i agree i don't like the wording - so more than happy to modify just not sure _to what_
They are different scopes is why they are different. 
You're adding a ebs encryption key which relates to a provider - not to a cluster. 
With the cluster_provider stuff, your referencing (maybe) the same provider, but it only relates to that specific cluster.

jordanbraiuka: yea that might be the best option

arjunrajshekhar:
> Based on my past PR experiences to TF, I would say this still needs example and test.

@h3nd24 - I think a test might be tricky since it requires a Run-In-Your-Own-Account account. Contributors outside of Instaclustr would have a hard time running those acceptance tests. Considering that, might be best to not include it? I do think an example would be good though.

Makefile
jordanbraiuka: Roger will address cheers

arjunrajshekhar (approved):

arjunrajshekhar:
Noting here that this PR is currently based on WIP changes to the API and should not be merged till the the changes have been deployed to Instaclustr's production environment.

README.md
```diff
@@ -233,12 +233,13 @@ Property | Description | Default
 key_id|Internal ID of the KMS encryption key. Can be found via GET to `https://api.instaclustr.com/provisioning/v1/encryption-keys`|""
 alias|KMS key alias, a human-readibly identifier specified alongside your KMS ARN|""
 arn|KMS ARN, identifier specifying provider, location and key in a ':' value seperated string|""
+key_provider|For customers running in their own account. Value specifying the the provider account’s name, similar to `instaclustr_cluster.cluster_provider.account_name`|INSTACLUSTR
```
alwyndavis, line 236: typo: "the the"

peter-toms (approved):

peter-toms:
adding a description to the PR would be good

arjunrajshekhar (approved):

PR #55, opened 2020-10-21
INS-10932

jordanbraiuka:
Fix issues with firewalls created with securitygroups.


arjunrajshekhar (approved):

arjunrajshekhar (changes requested):
Could the version be bumped in the makefile? Thanks!

h3nd24:
AFAICT the current test is only for the rule keyed by rule_cidr, should we add one based on security group id?

jordanbraiuka:
> AFAICT the current test is only for the rule keyed by rule_cidr, should we add one based on security group id?

The main issue with testing the security group id's is that it requires that you have already setup a vpc peer, and have a valid sg-id we can add. 
@arjunrajshekhar thoughts?

arjunrajshekhar (approved):

arjunrajshekhar (changes requested):
A placeholder firewall rule may be used to for acceptance testing, example - sg-fa3dE817

h3nd24 (dismissed):

arjunrajshekhar (approved):
Thanks for the changes. As per our discussion acceptance testing on security group based firewall rule is not needed here since it requires a legitimate security group to test with.

h3nd24 (approved):

jurgenweber:
I am receiving this error in the UI:

```
The following addresses failed to provision.
They have been removed from the list of allowed addresses.

sg-xxxxx
Re-add the addresses to re-attempt the provisioning operation.


```

jordanbraiuka:
Hey Jurgen, 

Is that Security Group in a Peered VPC? 
Firewall Rules for Security groups will only work correctly once your cluster has been VPC peered, and the security group resides in the VPC that you have peered.

Thanks

jurgenweber:
yes, I can copy and paste the same string in the UI and it works.

jurgenweber:
maybe it is a race condition sort of thing, the peer has not been accepted and finalised yet in the TF so it attempts it and fails.. but when I go and do it everything is in order.

jordanbraiuka:
Hi Jurgen, 
There is absolutely a race element to the function if you are performing the VPC peer in the same terraform state file. 

It's a good point though, i'll look to update some  information around performing this in one step, as it can be confusing. 

What i would suggest you try (and what has been working for me in my testing) is performing the operation with depends_on annotation. 

`
# request the vpc peer
resource "instaclustr_vpc_peering" "VPCPeering" {...}

# find the vpc in aws
data "aws_vpc_peering_connection" "clusterPeeringRequest" {
  depends_on           = [ instaclustr_vpc_peering.VPCPeering ]
...
}

# accept the vpc_peer in aws
resource "aws_vpc_peering_connection_accepter" "acceptPeeringRequest" {
  depends_on                = [ instaclustr_vpc_peering.VPCPeering ]
...}

# Add the security group to the firewall. 
resource "instaclustr_firewall_rule" "firewall_rules_for_handles" {
depends_on                = [ aws_vpc_peering_connection_accepter.acceptPeeringRequest ]

}

`

Let me know if that works for you, and in the background I will raise an internal ticket to get the documentation updated additionally.

jurgenweber:
Yeah, ok. That worked, with the dependencies set when I went to the settings page the sg rule was accepted/I had no error. Thank you!

PR #57, opened 2020-11-16
Password reset for schema reg and rest proxy

e12225:
This PR is to add the password reset functionality for Kafka Schema Registry and Kafka Rest Proxy in `ickafkascehma` and `ickafkarest` users of those respective bundles.


examples/main.tf
```diff
@@ -241,3 +241,57 @@ resource "instaclustr_cluster" "example-redis" {
...
+  }
+
+  bundle {
+    bundle = "KAFKA_SCHEMA_REGISTRY"
+    version = "5.0.0"
+  }
+}
+
+data "instaclustr_kafka_schema_registry_user_list" "kafka_schema_reg_user_list" {
+  cluster_id = "${instaclustr_clustr.example_kafka_schema_reg_and_reg_proxy.cluster_id}"
```
arjunrajshekhar, line 274: Doesn't look like `example_kafka_schema_reg_and_reg_proxy` matches the example resource you have defined. Also `instaclustr_clustr`

examples/main.tf
```diff
@@ -241,3 +241,57 @@ resource "instaclustr_cluster" "example-redis" {
...
+}
+
+data "instaclustr_kafka_rest_proxy_user_list" "kafka_rest_proxy_user_list" {
+  cluster_id = "${instaclustr_clustr.example_kafka_schema_reg_and_reg_proxy.cluster_id}"
+}
+
+resource "instaclustr_kafka_schema_registry_user" "example_user_test1" {
+  cluster_id = instaclustr_cluster.example_kafka_schema_reg_and_reg_proxy.cluster_id
```
arjunrajshekhar, line 282: Our examples are usually TF11 compatible, this should be changed to use string interpolation

examples/main.tf
```diff
@@ -241,3 +241,57 @@ resource "instaclustr_cluster" "example-redis" {
     }
   }
 }
+
```
arjunrajshekhar, line 244: Seeing a couple of issues when trying to plan and apply the following the new examples.

Example - 

`
Error: data.instaclustr_kafka_rest_proxy_user_list.kafka_rest_proxy_user_list: Provider doesn't support data source: instaclustr_kafka_rest_proxy_user_list

Error: data.instaclustr_kafka_schema_registry_user_list.kafka_schema_reg_user_list: Provider doesn't support data source: instaclustr_kafka_schema_registry_user_list
`

instaclustr/resource_kafka_schema_registry_user.go
```diff
@@ -0,0 +1,76 @@
...
+"encoding/json"
+"fmt"
+"log"
+
+"github.com/hashicorp/terraform/helper/schema"
+)
+
+func resourceKafkaSchemaRegistry() *schema.Resource {
+	return &schema.Resource{
+		Read:   resourceKafkaSchemaRegistryUserRead,
```
arjunrajshekhar, line 13: Resources should have an implementation for Create as well.

instaclustr/resource_kafka_schema_registry_user.go
```diff
@@ -0,0 +1,76 @@
+package instaclustr
```
arjunrajshekhar, line 1: As per the API - would it make sense to just create a generic resource and data source instead? Example - `instaclustr_bundle_user` ? That way we don't have to create all these separate resources that do the same thing.

arjunrajshekhar (changes requested):
Could the version on the makefile be bumped for this change?

examples/main.tf
e12225: yeah, I realised that I haven't added these two data sources to the provider.go. Will fix it.

instaclustr/resource_kafka_schema_registry_user.go
e12225: Will implement that. 👍

e12225: Done.

examples/main.tf
e12225: yeah. corrected it.

examples/main.tf
e12225: Refactored this as well.

examples/main.tf
e12225: Done

instaclustr/resource_kafka_schema_registry_user.go
e12225: Working on it at the moment

e12225: Implementation is done.

instaclustr/api_client.go
```diff
@@ -323,8 +343,8 @@ func (c *APIClient) UpdateKafkaUser(clusterID string, data []byte) error {
 	return nil
 }
 
-func (c *APIClient) DeleteKafkaUser(clusterID string, data []byte) error {
-	url := fmt.Sprintf("%s/provisioning/v1/%s/kafka/users", c.apiServerHostname, clusterID)
+func (c *APIClient) DeleteBundleUser(clusterID string, bundleName, data []byte) error {
```
arjunrajshekhar, line 346: `bundleName` should be a string

instaclustr/data_source_bundle_user_list.go
```diff
@@ -0,0 +1,49 @@
+package instaclustr
+
+import (
+	"fmt"
+	"log"
+
+	"github.com/hashicorp/terraform/helper/schema"
+)
+
+func dataSourceBundleUserList() *schema.Resource {
```
arjunrajshekhar, line 10: The readme will need to be updated for this resource and the data source

Makefile
```diff
@@ -1,5 +1,5 @@
 BIN_NAME="terraform-provider-instaclustr"
-VERSION=v1.7.0
+VERSION=v1.7.2
```
arjunrajshekhar, line 2: Seeing some merge conflicts, could you please resolve them?

instaclustr/resource_bundle_user.go
```diff
@@ -0,0 +1,166 @@
...
+				Required: true,
+			},
+
+			"username": {
+				Type:     schema.TypeString,
+				Required: true,
+				ForceNew: true,
+			},
+
+			"password": {
```
arjunrajshekhar, line 30: This should be marked as `Sensitive: true` as per [this change](https://github.com/instaclustr/terraform-provider-instaclustr/commit/b65d5b235c29165c9dbad7589271bb1407698889)

examples/main.tf
```diff
@@ -166,41 +169,43 @@ resource "instaclustr_cluster" "example-elasticsearch" {
...
+    bundle = "KAFKA_CONNECT"
+    version = "2.3.1"
+    options = {
+      target_kafka_cluster_id = "${instaclustr_cluster.example_kafka.id}"
+      vpc_id = "SEPARATE_VPC"
     }
+  }
 }
 
-resource "instaclustr_kafka_user" "kafka_user_charlie" {
```
arjunrajshekhar, line 194: Should retain the example for kafka user resource and data source

arjunrajshekhar (changes requested):

instaclustr/resource_bundle_user.go
```diff
@@ -0,0 +1,166 @@
...
+
+func resourceBundleUser() *schema.Resource {
+	return &schema.Resource{
+		Create: resourceBundleUserCreate,
+		Read:   resourceBundleUserRead,
+		Update: resourceBundleUserUpdate,
+		Delete: resourceBundleUserDelete,
+
+		Schema: map[string]*schema.Schema{
+			"cluster_id": {
```
arjunrajshekhar, line 19: All the properties that we don't expect to be updated should be marked as ForceNew. In this case - cluster_id, bundle_name should be marked.

instaclustr/resource_bundle_user.go
```diff
@@ -0,0 +1,166 @@
+package instaclustr
```
arjunrajshekhar, line 1: For code re-usability, I suggest moving the CRUD methods to another file and refactoring so that the new data source and resource + the kafka user resource and data sources re-use the CRUD code.

instaclustr/resource_bundle_user.go
```diff
@@ -0,0 +1,166 @@
...
+		},
+	}
+}
+
+func resourceBundleUserCreate(d *schema.ResourceData, meta interface{}) error {
+	cluster_id := d.Get("cluster_id").(string)
+	username := d.Get("username").(string)
+	bundle_name := d.Get("bundle_name").(string)
+
+	//todo: check if rest proxy or schema reg and then return nill, because those 2 bundles don't allow user creation
```
arjunrajshekhar, line 55: Does this need to be updated?

arjunrajshekhar (changes requested):

instaclustr/api_client.go
```diff
@@ -297,8 +297,8 @@ func (c *APIClient) DeleteEncryptionKey(keyID string) error {
 	return nil
 }
 
-func (c *APIClient) CreateKafkaUser(clusterID string, data []byte) error {
-	url := fmt.Sprintf("%s/provisioning/v1/%s/kafka/users", c.apiServerHostname, clusterID)
+func (c *APIClient) CreateBundleUser(clusterID string, bundleName string, data []byte) error {
+	url := fmt.Sprintf("%s/provisioning/v1/%s/%s/users", c.apiServerHostname, bundleName, clusterID)
```
arjunrajshekhar, line 301: The argument ordering here is incorrect

arjunrajshekhar (changes requested):

arjunrajshekhar:
Please see the following patch which contains changes for code re-use on the data sources and resources -

[patch.txt](https://github.com/instaclustr/terraform-provider-instaclustr/files/5572261/patch.txt)

instaclustr/api_client.go
e12225: resolved.

instaclustr/resource_bundle_user.go
e12225: done

examples/main.tf
e12225: added them back

instaclustr/resource_bundle_user.go
e12225: done

instaclustr/resource_bundle_user.go
e12225: This validation should be done for both creation & deletion of users of schema_registry and rest_proxy. Will add the validation and update the readme soon.

instaclustr/api_client.go
e12225: resolved

instaclustr/resource_bundle_user.go
e12225: done

e12225: updated the README as well.

instaclustr/data_source_bundle_user_list.go
e12225: updated

instaclustr/resource_cluster.go
```diff
@@ -281,6 +293,34 @@ func resourceCluster() *schema.Resource {
 					},
 				},
 			},
+			"kafka_rest_proxy_user_password": {
+				Type:     schema.TypeString,
+				Sensitive: true,
+				Optional: true,
+				ForceNew: false,
```
arjunrajshekhar, line 300: ForceNew is false by default so you may skip declaring it

instaclustr/resource_cluster.go
```diff
@@ -325,29 +365,165 @@ func resourceClusterCreate(d *schema.ResourceData, meta interface{}) error {
...
-	id, err := client.CreateCluster(jsonStr)
+	id, err := client.CreateCluster(jsonStrCreate)
 	if err != nil {
 		return formatCreateErrMsg(err)
 	}
 	d.SetId(id)
 	d.Set("cluster_id", id)
 	log.Printf("[INFO] Cluster %s has been created.", id)
-	return nil
+
```
arjunrajshekhar, line 381: This method is getting rather large, could it please be split out into smaller well named methods?

instaclustr/resource_cluster.go
```diff
@@ -325,29 +365,165 @@ func resourceClusterCreate(d *schema.ResourceData, meta interface{}) error {
...
+		}
+		if bundles[i].Bundle == "KAFKA_REST_PROXY" {
+			hasRestProxy = true
+		}
+		if bundles[i].Bundle == "KAFKA_SCHEMA_REGISTRY" {
+			hasSchemaRegistry = true
+		}
+	}
+
+	if len(waitForClusterState) > 0 {
```
arjunrajshekhar, line 403: Could return-early-return-often be used to simplify the logic here?

instaclustr/resource_cluster.go
```diff
@@ -325,29 +365,165 @@ func resourceClusterCreate(d *schema.ResourceData, meta interface{}) error {
...
+						}
+					}
+
+					return resource.NonRetryableError(resourceClusterRead(d, meta))
+				} else if !isKafkaCluster && (len(kafkaSchemaRegistryUserPassword) > 0 || len(kafkaRestProxyUserPassword) > 0) {
+					return resource.NonRetryableError(fmt.Errorf("[Error] Error updating the bundle user passwords, because it should be a KAFKA cluster in order to update the schema-registry or rest-proxy users"))
+				}
+			}
+
+			return resource.RetryableError(fmt.Errorf("[Error] Cluster is in state %s", cluster.ClusterStatus))
```
arjunrajshekhar, line 465: Suggest to change the error string here to -

`fmt.Errorf("[DEBUG] Cluster is in state %s, waiting for it to reach state %s", cluster.ClusterStatus, waitForClusterState)`

instaclustr/resource_cluster.go
arjunrajshekhar, line 403: There should be some validation included to enforce waiting for the cluster to be running when the passwords are provided.

arjunrajshekhar (changes requested):

arjunrajshekhar:
With the latest changes I'm seeing an issue with terraform plan -
1. Create a cluster. Example, terraform apply with the following config - 
```
resource "instaclustr_cluster" "example_kafka_bundle_user_update" {
  cluster_name = "test_kafka"
  node_size = "r5.large-500-gp2"
  data_centre = "US_WEST_2"
  sla_tier = "NON_PRODUCTION"
  cluster_network = "192.168.0.0/18"
  cluster_provider = {
    name = "AWS_VPC"
  }
  rack_allocation = {
    number_of_racks = 3
    nodes_per_rack = 1
  }
  bundle {
    bundle = "KAFKA"
    version = "2.5.1"
    options = {
      auth_n_authz = true
      dedicated_zookeeper = true
      zookeeper_node_size = "zk-production-m5.large-60"
      zookeeper_node_count = 3
    }
  }
  bundle {
    bundle = "KAFKA_REST_PROXY"
    version = "5.0.0"
  }
  bundle {
    bundle = "KAFKA_SCHEMA_REGISTRY"
    version = "5.0.0"
  }
}
```
2. Run terraform plan <- shows that there are changes. The expected outcome is that there should be no changes

instaclustr/resource_cluster.go
```diff
@@ -325,29 +365,165 @@ func resourceClusterCreate(d *schema.ResourceData, meta interface{}) error {
...
+		})
+	} else {
+		return resourceClusterRead(d, meta)
+	}
 }
 
 func resourceClusterUpdate(d *schema.ResourceData, meta interface{}) error {
 	d.Partial(true)
-	// currently only cluster resize is supported
-	if !d.HasChange("node_size") {
```
arjunrajshekhar, line 347: We should be either be throwing an error for all the properties that we don't support update for, or add ForceNew: true to all those non-updatable properties (ForceNew is the preferable way). Otherwise, this can lead to issues where after updating a non-updatable property, `terraform plan` will always suggest that there is a change to the resource.

instaclustr/resource_cluster.go
```diff
@@ -325,28 +366,151 @@ func resourceClusterCreate(d *schema.ResourceData, meta interface{}) error {
...
+	if err != nil {
+		log.Printf("[ERROR] Error creating the bundle user update request : %s", err)
+		return nil
+	} else {
+		return jsonStrUpdateBundleUser
 	}
+}
+
+func checkForBundleAvailability(bundles []Bundle) {
+	for i := 0; i < len(bundles); i++ {
```
h3nd24, line 499: optionally, this can be defined as ```
for _, b := range bundles { 
  if (b.Bundle == "KAFKA") ...
```

File: instaclustr/structs.go [hunk 21]
```diff
@@ -158,6 +158,21 @@ type EncryptionKey struct {
 	Provider string `json:"provider,omitempty"`
 }
 
+type CreateBundleUserRequest struct {
```
arjunrajshekhar, line 161: Looks like these three structs have been defined but are unused

instaclustr/resource_cluster.go
```diff
@@ -145,10 +172,12 @@ func resourceCluster() *schema.Resource {
...
 							Required: true,
+							ForceNew: true,
 						},
 						"version": {
 							Type:     schema.TypeString,
 							Required: true,
+							ForceNew: true,
 						},
 						"options": {
 							Type:     schema.TypeMap,
```
EngineersBox, line 183: Is there any reason we don't include `ForceNew` on this composite object? Same question applies to others of similar situation.

EngineersBox (changes requested):
Some comments regarding code quality, maintainability and general implementation questions.

README.md
```diff
@@ -112,6 +112,11 @@ pci_compliant_cluster|Accepts true/false. Creates the cluster with PCI complianc
 cluster_provider|The information of infrastructure provider. See below for its properties.|Required
 rack_allocation|The number of resources to use. See below for its properties.|Optional, but Required for all Bundle types excluding Redis.
 bundle|Array of bundle information. See below for its properties.|Required
+kafka_rest_proxy_user_password|The password of kafka rest proxy bundle user, if it's a Kafka cluster with rest-proxy addon|Optional
+kafka_schema_registry_user_password|The password of kafka schema registry bundle user, if it's a Kafka cluster with schema-registry addon|Optional
```
arjunrajshekhar, line 116: Would be best to mention the dependency on waiting for the state in the README

arjunrajshekhar (changes requested):

arjunrajshekhar (approved):

EngineersBox (approved):

h3nd24 (approved):

PR #60, opened 2020-12-03
Introducing generic resources and data sources

e12225:
This PR is to include generics user resources and data sources instead of the bundle specific user resources and data sources.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,5 +1,5 @@
 BIN_NAME="terraform-provider-instaclustr"
-VERSION=v1.7.0
+VERSION=v1.7.3
 
 .PHONY: install clean all build test testacc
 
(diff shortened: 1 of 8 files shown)


h3nd24 (changes requested):
Something that came to mind on this whole bundle user supposed to be generic but KSR and RP needs to attach the info into the cluster, I would think we'll need some sort of validation to prevent determined user from creating a KSR and RP bundle user resource, despite all the warnings and README. Because if they successfully created a KSR bundle user resource, we are faced with the original problem as to why we need KSR and RP info in the cluster.

peter-toms:
Closing this item as internally we are taking a different angle to solve this problem.

Issue #32, opened 2020-06-18
Allow terraform import


renaudhager:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or other comments that do not add relevant new information or questions, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Description

It could be very useful to leverage the `import` feature offer by Terraform. 
The idea is that any instaclustr resources that has been manually created can imported in the Terraform state, using `terraform import` command.

peter-toms:
thanks for the suggestion @renaudhager ill raise it with the team and get back to you

ProfMoo:
This would be HUGE for our workflow. Currently, we have a cluster defined in Terraform, but the name given to the Terraform resource is incorrect. AFAIK, there isn't a way to change that without deleting the resource and recreating. `import` could help out here.

slater-ben:
INS-8583 corresponding Instaclustr internal issue

Chrismelba-Instaclustr:
This has been added with v1.8. Thanks everyone for the suggestion!

PR #61, opened 2020-12-31
Ins 8583

wjjmjh:
Resolves issues #32
Internal reference - INS-8583


wjjmjh:
@peter-toms @EngineersBox 

Hi team! please have a look when you have time, cheers!

EngineersBox (commented):
Did an initial pass at the PR, will do another one.

test/import_resource_test.go
```diff
@@ -0,0 +1,209 @@
...
+	"fmt"
+	"github.com/hashicorp/terraform/helper/resource"
+	"github.com/hashicorp/terraform/terraform"
+	"github.com/instaclustr/terraform-provider-instaclustr/instaclustr"
+	"io/ioutil"
+	"os"
+	"testing"
+)
+
+func TestAccCluster_importBasic(t *testing.T) {
```
EngineersBox, line 13: We should have multiple test files for each of the offerings (C*, Kafka, Redis) and do tests for each one as they use different bundle configurations and bundle specific schema sections. These should be covered by tests.

test/data/valid_single_bundle.tf
```diff
@@ -0,0 +1,33 @@
...
+    name = "AWS_VPC"
+  }
+  rack_allocation = {
+    number_of_racks = 3
+    nodes_per_rack = 1
+  }
+
+  bundle {
+    bundle = "APACHE_CASSANDRA"
+    version = "4.0"
```
EngineersBox, line 25: C* version shouldn't be a preview, instead probably 3.11.8, or some equivalent therein

test/data/valid_single_bundle.tf
```diff
@@ -0,0 +1,33 @@
...
+  sla_tier = "NON_PRODUCTION"
+  cluster_network = "192.168.0.0/18"
+  private_network_cluster = false
+  pci_compliant_cluster = false
+  cluster_provider = {
+    name = "AWS_VPC"
+  }
+  rack_allocation = {
+    number_of_racks = 3
+    nodes_per_rack = 1
```
EngineersBox, line 20: A non-default rack allocation would be good to have instead. This would give you confidence that it handles not just default behaviour but also custom values E.g: 4 racks and 2 nodes per rack

instaclustr/resource_vpc_peering.go
```diff
@@ -139,9 +144,7 @@ func resourceVpcPeeringRead(d *schema.ResourceData, meta interface{}) error {
 	d.Set("peer_vpc_id", vpcPeering.PeerVpcID)
 	d.Set("peer_account_id", vpcPeering.PeerAccountID)
 	d.Set("aws_vpc_connection_id", vpcPeering.AWSVpcConnectionID)
-
-	before, _ := d.GetChange("peer_subnet")
-	d.Set("peer_subnet", before)
+	d.Set("peer_subnet", vpcPeering.PeerSubnet)
```
EngineersBox, line 147: Curious about this change, i haven't look to deeply yet. But what was the reason behind it?

EngineersBox (commented):
Added a couple of comments on structure, testing and a change to vpc peering reading

instaclustr/resource_vpc_peering.go
wjjmjh: Good question, when importing resources, the ResourceData would be initialised with only ID set, so before in "before, _ := d.GetChange("peer_subnet")" will always be nil when importing;

Set would be more applicable for resourceRead.

test/data/valid_single_bundle.tf
wjjmjh: 4 racks not accepted as we should pick one from [1, 3, 5];

so I updated it to 

number_of_racks = 5
 nodes_per_rack = 2

instaclustr/resource_vpc_peering.go
EngineersBox: Sounds reasonable to me 👍🏼

instaclustr/resource_cluster.go
```diff
@@ -26,6 +26,11 @@ func resourceCluster() *schema.Resource {
 		Read:   resourceClusterRead,
 		Update: resourceClusterUpdate,
 		Delete: resourceClusterDelete,
+
+		Importer: &schema.ResourceImporter{
+			State: resourceClusterStateImport,
+		},
+
```
EngineersBox, line 33: The API response doesn't fully support all of the required fields/structure needed for importing clusters. Namely missing support for Kafka, Redis, etc.

EngineersBox (commented):
Just added a note about cluster importing, notably as a result of internal conversion in Instaclustr

PR #64, opened 2021-02-04
Added support for alpine linux via CGO_ENABLED

toddschilling:


.circleci/config.yml
```diff
@@ -30,6 +30,8 @@ jobs:
             BIN_NAME="terraform-provider-instaclustr"
             FULL_BIN_NAME="${BIN_NAME}_${CIRCLE_TAG}"
             SHASUM_NAME="${BIN_NAME}_${CIRCLE_TAG:1}"
+            env GOOS=linux GOARCH=amd64 CGO_ENABLED=0 make build VERSION=${CIRCLE_TAG}
```
Jack-Kilrain, line 33: **[Question]** Have seen a few [StackOverflow posts](https://stackoverflow.com/questions/36279253/go-compiled-binary-wont-run-in-an-alpine-docker-container-on-ubuntu-host) about issues with go binaries on alpine with the net package (we use that for the api interactions). I don't know that will be an issue as the binary is a plugin, but depending on terraform's instantiation of plugins it might cause issues

The suggested fix on posts is to use the `netgo` inbuilt package instead of the default net package. This would be a change to the build command:
```bash
go build -tags netgo -a -v -o bin/$(BIN_NAME)_$(VERSION) main.go
```

Jack-Kilrain (commented):
Added a comment about a potential issue with the compiled binary on alpine images

.circleci/config.yml
toddschilling: You're correct.  Let me test this locally to verify and I'll update the PR once I've confirmed that this works.

toddschilling: @Jack-Kilrain I've made the suggested update.  Much better!

Jack-Kilrain: Thanks for testing that, much appreciated!

Jack-Kilrain (approved):

Chrismelba-Instaclustr:
Looks Good,

We're about to push an update to support importing, so will roll this through then.

pickgr (approved):

Chrismelba-Instaclustr (approved):

PR #71, opened 2021-03-10
Enable auto test on Circle CI

iamhail:
Internal reference - INS-12822


instaclustr/resource_cluster.go
```diff
@@ -725,6 +709,64 @@ func resourceClusterRead(d *schema.ResourceData, meta interface{}) error{
...
+
+func processProvider(d *schema.ResourceData, data map[string]interface{}) (newData map[string]interface{}) {
+	// This is used to ignore values that are not set in the resource
+	// Otherwise terraform will store the default and generated values (e.g. Provider Account Name)
+	// into the state and result a diff between the state and resource (plan)
+	newData = make(map[string]interface{})
+	for k, v := range data {
+		resource := d.Get(fmt.Sprintf("cluster_provider.%s", k))
+		// Store the required field "name"
+		// There doesn't seem to be a way to programmatically get the "required" schemas with ResourceData
```
Chrismelba-Instaclustr, line 761: [requested] Perhaps we could store a list of requiredResources and iterate through them? Might make it more obvious in future how to add a required resource to this check

instaclustr/resource_cluster.go
```diff
@@ -725,6 +709,64 @@ func resourceClusterRead(d *schema.ResourceData, meta interface{}) error{
...
+			data[k] = fmt.Sprintf("%v", reflect.Indirect(valueOfV.Elem()).Interface())
+		} else {
+			data[k] = fmt.Sprintf("%v", valueOfV)
+		}
+	}
+
+	return data
+}
+
+func processProvider(d *schema.ResourceData, data map[string]interface{}) (newData map[string]interface{}) {
```
Chrismelba-Instaclustr, line 753: Is this only useful for processing the provider? Or could it be used more generally for other string fields?

test/test_utils.go
```diff
@@ -2,84 +2,55 @@ package test
...
-			"IC_SSL_ENABLED_PROTOCOLS", "IC_SSL_TRUSTSTORE_PASSWORD", "IC_SSL_PROTOCOL", "IC_SECURITY_PROTOCOL",
-			"IC_SASL_MECHANISM", "IC_SASL_JAAS_CONFIG", "IC_BOOTSTRAP_SERVERS", "IC_TRUSTSTORE"}
-		for _, s := range env_vars {
-			if v := os.Getenv(s); v == "" {
-				fatalMessage := fmt.Sprintf("When IC_TEST_KAFKA_CONNECT is set, %s must be set for acceptance tests", s)
-				t.Fatal(fatalMessage, s)
-			}
-		}
+func getOptionalEnv(key, fallback string) string {
+	value := os.Getenv(key)
```
EngineersBox, line 13: **[Question]** If the env var isn't set, will this not return a nil pointer?

instaclustr/resource_cluster.go
```diff
@@ -628,35 +628,17 @@ func resourceClusterRead(d *schema.ResourceData, meta interface{}) error{
...
 	d.Set("cluster_id", cluster.ID)
 	d.Set("cluster_name", cluster.ClusterName)
-	clusterProvider := make(map[string]interface{}, 0)
-	clusterProvider["name"] = cluster.DataCentres[0].Provider
-	d.Set("cluster_provider", clusterProvider)
 
-	baseBundle := make(map[string]interface{}, 0)
-	baseBundle["bundle"] = cluster.BundleType
+	clusterProvider := make(map[string]interface{}, 0)
+	mapstructure.Decode(cluster.Provider, &clusterProvider)
```
EngineersBox, line 633: **[Question]** Any particular reason to use fixed/hard decode over `WeakDecode()`? Surprised there was no issues with type mismatching.

instaclustr/resource_cluster.go
```diff
@@ -725,6 +709,64 @@ func resourceClusterRead(d *schema.ResourceData, meta interface{}) error{
 	return nil
 }
 
+func getBundlesFromCluster(cluster *Cluster) ([]map[string]interface{}, error) {
+	baseBundle := make(map[string]interface{}, 0)
```
EngineersBox, line 713: **[Required, Memory Usage]** This map will always have 3 entries, since there are no loop based assignments. They are as follows:
```go
baseBundle["bundle"] = cluster.BundleType
baseBundle["options"] = convertedBundleOptions
baseBundle["version"] = cluster.BundleVersion
```

EngineersBox (changes requested):
Some changes desired, mostly just comments and questions

instaclustr/resource_cluster.go
iamhail: Tried that before from Arjun's suggestion but didn't work

instaclustr/resource_cluster.go
iamhail: I feel like this is still kind of a hacky solution and we should try to avoid using it

instaclustr/resource_cluster.go
iamhail: As commented above this feels like a hacky solution and creating a list of requiredResources would add a duplication to our resource data that already exists so when we need to add more "required resource" check we should be attempting to do it differently (i.e. shouldn't encourage this solution)

test/data/valid_with_custom_vpc.tf
```diff
@@ -23,6 +23,13 @@ resource "instaclustr_cluster" "vpc_cluster" {
...
     bundle {
         bundle = "APACHE_CASSANDRA"
-        version = "3.11.4"
+        version = "apache-cassandra-3.11.8.ic2"
+        options = {
+            auth_n_authz = false
+            client_encryption = false
+            continuous_backup_enabled = false
+            lucene_enabled = false
+            use_private_broadcast_rpc_address = false
```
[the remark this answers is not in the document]
hail-insta: It would, but I'd prefer not to change the scope of the tests as part of this (just focusing on fixing them)

test/test_utils.go
hail-insta: Tested with no env var set, it will pass the if condition and return the fallback value (used to get the API url and the default fallback is "api.instaclustr.com"

Chrismelba-Instaclustr (approved):

Makefile
```diff
@@ -38,6 +39,48 @@ ifndef IC_PROV_VPC_ID
...
+	@echo "TARGET for test name must be set for running specific tests"
+	@exit 1
+endif
+
+ifeq (,$(findstring Test, $(TARGET)))
+	@echo "The target is not a test"
+	@exit 1
+endif
+
+ifneq (,$(findstring Acc, $(TARGET)))
```
arjunrajshekhar, line 53: [Desired, maintainability] - so that we don't rely on naming patterns, would it make sense to define these checks on the target tests?

examples/main.tf
```diff
@@ -49,7 +49,7 @@ resource "instaclustr_cluster" "example" {
 
   bundle {
     bundle = "SPARK"
-    version = "apache-spark:2.3.2"
+    version = "2.3.2"
```
arjunrajshekhar, line 52: [Question] - Pretty sure this was working before. Was there a breaking API change that requires this?

Makefile
```diff
@@ -1,7 +1,8 @@
+
 BIN_NAME="terraform-provider-instaclustr"
 VERSION=v1.8.0
```
arjunrajshekhar, line 3: It appears that there may be some material change for users of the terraform provider (example - the import looks like it has been updated), could we bump up the version?

test/import_resource_test.go
```diff
@@ -171,7 +168,7 @@ func TestKafkaUserResource_importBasic(t *testing.T) {
 				Config: createClusterConfig,
 				Check: resource.ComposeTestCheckFunc(
 					testCheckResourceValidKafka("instaclustr_cluster.kafka_cluster"),
-					checkKafkaClusterRunning(hostname, username, apiKey),
+					checkClusterRunning("kafka_cluster", hostname, username, apiKey),
```
arjunrajshekhar, line 171: Excellent add 🏆

Makefile
iamhail: good point, will bump

examples/main.tf
iamhail: No, this is so we don't encourage usages of the internal versions (we try to avoid that now)

But in the tests we had to change to use the internal versions because some tests check against the plan and because the API response from terraform-description endpoint returns the ic version, it detects diff and fails the tests

Makefile
iamhail: Had a discussion of the checks on the target tests and decided that it is creating some duplication especially on specific tests like kafka connect.
We shouldn't be running target tests in the first place imo (tests should run as a whole, and especially now since we are working to make them run automatically during PR phase), but decided to just provide this so we can run tests during development in some way. What we want is that when we run tests, we should just run either `make test` or `make testacc`

examples/main.tf
arjunrajshekhar: Ah got it, thanks for confirming. I'm guessing that means terraform import won't work with internal versions then. Either way all good.

Makefile
arjunrajshekhar: Yeah there would certainly be a bit of duplication by getting the tests to sort out their own dependencies. It's too bad golang doesn't support annotations because it would have made it easier to mark and check simple test dependencies with something like that. Happy about keeping this as is.

arjunrajshekhar (dismissed):

arjunrajshekhar (approved):

Chrismelba-Instaclustr (approved):

PR #73, opened 2021-03-12
Zookeeper cluster support

h3nd24:
Support for ZooKeeper Cluster


test/data/valid_apache_zookeeper.tf
```diff
@@ -0,0 +1,31 @@
...
+    rack_allocation = {
+        number_of_racks = 3
+        nodes_per_rack = 1
+    }
+    
+    bundle {
+        bundle = "APACHE_ZOOKEEPER"
+        version = "apache-zookeeper:3.5.8"
+        options = {
+            zookeeper_node_count = 3
```
EngineersBox, line 28: **[Question]** This doesn't feel right, since this is a main offering the node count should be calculated through the rack allocation. Is there something different here that I'm missing?

EngineersBox (changes requested):
Looking good, but some things to address on the test case and also the node count

test/data/valid_apache_zookeeper.tf
h3nd24: So for this bundle it is similar to REDIS in a sense that we don't need the rack allocation. The rack allocation is there because the version I'm PR-ing against still have a problem whereby rack allocation is "expected", i.e., we can create a zookeeper or Redis cluster without rack allocation, but then the cluster read still returns rack allocation, which makes terraform think there is a need for an update where it shouldn't be.

EngineersBox: Alrighty, that makes sense. Terraform's deadly self-update cycle can be a pain. I suppose it's something that needs to be addressed in future with the structuring of the resources.

hail-insta (approved):

Jack-Kilrain (dismissed):

EngineersBox (approved):

Jack-Kilrain (approved):

tharinduamila-insta (approved):
LGTM

tharinduamila-insta (approved):

tharinduamila-insta (approved):

hail-insta (approved):

tharinduamila-insta (approved):

hail-insta (dismissed):

hail-insta (approved):

tharinduamila-insta (approved):

PR #75, opened 2021-03-23
Ins 12957 update api structure

Chrismelba-Instaclustr:
Updating the go structure to match the new API in order to support multiple add-on bundles per cluster


instaclustr/structs.go
```diff
@@ -95,15 +95,15 @@ type Cluster struct {
...
+	AddonBundles			   []map[string]interface{}	 `json:"addonBundles"`
 	Username                   string        	`json:"username"`
 	InstaclustrUserPassword    string        	`json:"instaclustrUserPassword"`
 	SlaTier                    string        	`json:"slaTier"`
 	ClusterCertificateDownload string        	`json:"clusterCertificateDownload"`
 	PciCompliance              string        	`json:"pciCompliance"`
 	BundleOption               *BundleOptions 	`json:"bundleOptions"`
 	DataCentres                []DataCentre  	`json:"dataCentres"`
-	Provider				   ClusterProvider	`json:"clusterProvider"`
+	Provider				   []ClusterProvider	`json:"clusterProvider"`
```
Jack-Kilrain, line 106: **[Question]** I don't understand why there would be more than one `ClusterProvider`. Could you clarify this for me please?

instaclustr/resource_cluster.go
```diff
@@ -630,7 +630,7 @@ func resourceClusterRead(d *schema.ResourceData, meta interface{}) error{
 	d.Set("cluster_name", cluster.ClusterName)
 
 	clusterProvider := make(map[string]interface{}, 0)
-	mapstructure.Decode(cluster.Provider, &clusterProvider)
+	mapstructure.Decode(cluster.Provider[0], &clusterProvider)
```
Jack-Kilrain, line 633: **[Question]** Why would we support multiple providers if it's only the first that is being used?

Jack-Kilrain (changes requested):
Got some questions about the array fo `ClusterProvider` other than that, looks pretty tidy

instaclustr/structs.go
Jack-Kilrain: Talked to Micheal and we do support multiple providers for a single cluster. Though will need to change the state updater in read to handle this instead of just the first element.

instaclustr/resource_cluster.go
Jack-Kilrain: Will need to update this to use all of the providers from the list not just the first one

Jack-Kilrain: Talked with @Chrismelba-Instaclustr on this, we will keep this as is since the TF provider won't support multiple providers in it's current state.

Chrismelba-Instaclustr: As discussed, console supports multi-provider provisioning, but API currently doesn't. This structure allows the API to return the information for a multi-provider cluster, but terraform will just take the first instance returned since it doesn't support multi-provider provisioning (thus there will only be one provider returned in the list)

Jack-Kilrain (approved):
LGTM

Jenkinsfile
```diff
@@ -0,0 +1,17 @@
+pipeline {
+  agent any
+  stages {
+    stage('merge') {
+      steps {
+        git(url: 'https://github.com/Chrismelba-Instaclustr/terraform-provider-instaclustr/', branch: '{Branch-SHA}')
```
peter-toms, line 6: [required] shouldn't that be the main instaclustr branch ?

peter-toms (changes requested):

peter-toms (dismissed):

Makefile
```diff
@@ -1,6 +1,10 @@
 
 BIN_NAME="terraform-provider-instaclustr"
-VERSION=v1.9.1
+
+VERSION=v1.9.2
+release_version:
```
hail-insta, line 5: what's this target for?
also it doesn't represent a target, should make it .PHONY

Chrismelba-Instaclustr: Not actually part of this ticket, but it lets us get the version from the makefile for the jenkins tagging. What's a .PHONY do?

hail-insta: Makefile targets represent a file to determine "cached builds" - if you `make stuff` and `stuff` file exists then make won't run that target.
Phony says that the target is not a real target (it doesn't create a file with `release_version`), meaning that it will run the target regardless of the presence of "cached file". release_version being a real target will prevent it to run (echo wont happen) if `release_version` file exists. of course that's not going to happen and it's not really a concern, but we should follow the standard

EngineersBox (approved):
LGTM

Jack-Kilrain (approved):
LGTM

hail-insta (approved):

Issue #63, opened 2021-02-04
Unable to execute on Alpine 3.12


pickgr:
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or other comments that do not add relevant new information or questions, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Terraform Version

terraform 12.28

### Affected Resource(s)

* instaclustr_1.7.3

### Debug Output

exit status 1: running "tf-plan dev" in "/home/atlantis/.atlantis/repos/seismic/terraform/1280/default/providers/instaclustr/azure-dev/westus/dev-eventbus": 

Error: Failed to instantiate provider "instaclustr" to obtain schema: fork/exec /home/atlantis/.terraform.d/plugins/linux_amd64/terraform-provider-instaclustr_v1.7.3: no such file or directory

No workflow (dev) tfvars file to include.
Loading the workspace (default) tfvars file!

### Expected Behavior

terraform init should instantiate the provider

### Actual Behavior

See the error above.

### Steps to Reproduce

This running on an Alpine 3.12 container.  Specifically, we use the open source atantis project to run our terraform

https://github.com/runatlantis/atlantis/blob/master/docker-base/Dockerfile#L4

1. `terraform init`

### Important Factoids

We have verified that the binary is installed in the correct place even though the error says "no such file or directory"

The binary we are trying to use is from 

https://github.com/instaclustr/terraform-provider-instaclustr/releases/download/v1.7.3/terraform-provider-instaclustr_v1.7.3_linux_amd64.zip

We have 2 other plugins (ibm and kafka) that run fine.  Note that only instaclustr shows dynamic linking below"

```
terraform-provider-ibm_v1.20.1:        ELF 64-bit LSB executable, x86-64, version 1 (SYSV), statically linked, Go BuildID=GCMn5HhuuXZOqpY21QlD/_L0vEtS3sXKKnsXKOFG4/j1FUonfQVGaR1e8uBwwO/-eepCW4-Kjoi2Vvs-3PH, stripped
[...]

PR #77, opened 2021-04-07
Fix for Issue #63

toddschilling:
I've verified that the build generated by CircleCI using the `circleci/golang:1.14` and the `-tags netgo -a -v` parameters does not work on the `runatlantis/atlantis` image.  I also verified that adding `CGO_ENABLED=0` fixes it. 

We are currently running `terraform-provider-instaclustr` with `CGO_ENABLED=0` on our CICD platform and it has work for all operations thus far.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -30,7 +30,7 @@ jobs:
             BIN_NAME="terraform-provider-instaclustr"
             FULL_BIN_NAME="${BIN_NAME}_${CIRCLE_TAG}"
             SHASUM_NAME="${BIN_NAME}_${CIRCLE_TAG:1}"
-            env GOOS=linux GOARCH=amd64 make build FLAGS="-tags netgo -a -v" VERSION=${CIRCLE_TAG}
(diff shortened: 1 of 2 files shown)


pickgr (approved):

peter-toms:
Thanks so much @toddschilling!  For visibility we will be testing this change alongside some enhancements we have in this space (estimating sometime in the next couple of weeks).  Once verified will give it the green light from here.  Cheers!

hail-insta:
Thanks @toddschilling! We have tested and verified this change. We will just bump the version in the PR and take it through to deployment soon.

hail-insta (approved):

hail-insta:
Updating PR as per comment in the issue #63

> After some internal discussions, we have decided to build the standard linux binary with static linking to include required packages within the binary to support compatibility. This means that the alpine and linux builds will be identical and the alpine build will still remain in our release, but we will likely deprecate it in the future.
Thanks @toddschilling - we will still use your change for the build. Appreciate the suggestion on stripping and DWARF debug removal as it has significantly reduced the binary size

hail-insta (approved):

peter-toms (approved):

PR #81, opened 2021-04-15
Run tests from Circle CI upon PR commits

hail-insta:


Makefile
```diff
@@ -17,7 +17,7 @@ build:
 	go build $(FLAGS) -o bin/$(BIN_NAME)_$(VERSION) main.go
 
 test:
-	cd test && go test -v -timeout 120m -count=1
+	go test -v -timeout 120m -count=1 ./...
```
EngineersBox, line 20: **[Question]** Not to sure what the intend of the the directory changes here are, seems like its going up two levels from the current context. Could I grab some clarification on this?

instaclustr/api_mock_client.go
```diff
@@ -1,14 +1,12 @@
...
 import (
 	"net/http"
 	"time"
-
-	"github.com/instaclustr/terraform-provider-instaclustr/instaclustr"
 )
 
 type APIMockClient struct {
-	instaclustr.APIClient
+	APIClient
```
EngineersBox, line 9: **[Desired, Optimisation]** Currently, this will pass the `APIClient` struct on the heap and store it in an instance of this struct. That's not great. Instead, we should pass a pointer to an APIClient to this structure as that avoids stack memory restructuring and also eliminates the overhead with memory drop with context changes

instaclustr/resource_cluster.go
```diff
@@ -726,9 +727,13 @@ func getBundlesFromCluster(cluster *Cluster) ([]map[string]interface{}, error) {
...
-	if cluster.AddonBundles != nil {
-		for _, addonBundle := range cluster.AddonBundles{
-			bundles = append(bundles, addonBundle)	
+	addonBundles := cluster.AddonBundles
+	sort.Slice(addonBundles, func(i, j int) bool {return addonBundles[i]["bundle"].(string) > addonBundles[j]["bundle"].(string)})
+	if addonBundles != nil {
+		for _, addonBundle := range addonBundles {
+			if len(addonBundle) != 0 {
+				bundles = append(bundles, addonBundle)
+			}
```
EngineersBox, line 736: **[Desired, Maintainability]** Would be good to invert the conditional here, would make this more logical and maintainable:
```go
if addonBundles != nil {
    return nil, nil
}
for _, addonBundle := range addonBundles {
    if len(addonBundle) != 0 {
        bundles = append(bundles, addonBundle)
    }
}
```

instaclustr/resource_cluster_unit_test.go
```diff
@@ -0,0 +1,99 @@
...
+func TestCreateBundleUserUpdateRequest(t *testing.T) {
+	testUsername := "hail"
+	testPassword := "reallySecure123"
+	testBundleRequest := createBundleUserUpdateRequest(testUsername, testPassword)
+
+	expectedOutput := fmt.Sprintf("{\"username\":\"%s\",\"password\":\"%s\"}", testUsername, testPassword)
+
+	if fmt.Sprintf("%s", testBundleRequest) != expectedOutput {
+		t.Fatalf("Incorrect request returned.\nExpected:%s\nActual:%s", expectedOutput, testBundleRequest)
+	}
```
EngineersBox, line 18: **[Desired, Maintainability]** It doesn't feel right that we are comparing manually stringified JSON. I would prefer to see it done on a bytes marshalled basis:
```go
func JSONBytesEqual(a, b []byte) (bool, error) {
    var j, j2 interface{}
    if err := json.Unmarshal(a, &j); err != nil {
        return false, err
    }
    if err := json.Unmarshal(b, &j2); err != nil {
        return false, err
    }
    return reflect.DeepEqual(j2, j), nil
}

func main() {
    b1 := []byte(`{"x": {"t": 1, "s": 2}, "z": 1}`)
    b2 := []byte(`{"z": 1, "x": {"s": 2, "t": 1}}`)

    eq, err := JSONBytesEqual(b1, b2) // false
    eq, err = JSONBytesEqual(b1, b1) // true
}
```

EngineersBox (changes requested):
Looking good, some go-specific related suggestions for maintainability

Makefile
hail-insta: This will run every test in all directories and sub directories. Previously we only had tests in `test` folder. Now we have two levels of tests - acceptance tests in `acc_tests` folder and unit tests in `instaclustr` folder (this is because unit tests are used to test the internal methods used for CRUD but they are unexported so the tests need to be in the same package). Essentially, `./...` will run every test in the repository.

EngineersBox: Ah ok, gotcha. Sounds good.

instaclustr/resource_cluster_unit_test.go
hail-insta: addressed - comparing the bytes slice so we dont need any data processing for test

instaclustr/api_mock_client.go
EngineersBox: Discussed with Hail, will remove the APIMockClient struct and use a function to return a mocked API client

EngineersBox (changes requested):

peter-toms (approved):

.circleci/config.yml
```diff
@@ -67,14 +94,61 @@ jobs:
             -n "Release ${CIRCLE_TAG}" \
             -b "Generated by Instaclustr's Circle CI account." \
             ${CIRCLE_TAG} ./bin/
-
+jobs:
+  build:
+    docker:
+      - image: circleci/golang:1.14
```
peter-toms, line 100: (question) any reason why we are not building on go 1.16 ?

README.md
```diff
@@ -1,12 +1,17 @@
 # Terraform Instaclustr Provider
 
-A [Terraform](http://terraform.io) provider for managing Instaclustr Platform resources.  
+[![CircleCI](https://img.shields.io/circleci/build/gh/instaclustr/terraform-provider-instaclustr/master?style=for-the-badge)](https://app.circleci.com/pipelines/github/instaclustr/terraform-provider-instaclustr)
```
peter-toms, line 3: (question) badge is still showing build as failing ?  On release we'd expect a passing badge

peter-toms (approved):

.circleci/config.yml
hail-insta: This is what we have been using, and I want to avoid opportunistically upgrading the version and add unnecessary overhead given that this PR was already broken by number of other things

README.md
hail-insta: The badge is showing failing because it doesn't reflect builds of tags. Our releases are based off tags, so nothing runs on master. On merge of this PR however it will run the tests on master branch which will be reflected on the badge

Chrismelba-Instaclustr (approved):
LGTM!

peter-toms (approved):

PR #83, opened 2021-04-19
Ins 12901 private contact point is null

Chrismelba-Instaclustr:
Fixing a bug where private and public contact points wouldn't populate


Makefile
```diff
@@ -1,20 +1,22 @@
 
 BIN_NAME="terraform-provider-instaclustr"
 
-VERSION=v1.9.5
+VERSION=1.9.6
+INSTALL_FOLDER=$(HOME)/.terraform.d/plugins/example.com/instaclustr/instaclustr/$(VERSION)/darwin_amd64
```
peter-toms, line 5: [required] does that need to be around ? example.com ? feels like some testing ?

instaclustr/resource_cluster.go
```diff
@@ -668,14 +676,13 @@ func resourceClusterRead(d *schema.ResourceData, meta interface{}) error{
 		}
 	}
 	rackCount := len(rackList)
-	nodesPerRack := nodeCount/rackCount
+	nodesPerRack := nodeCount / rackCount
```
peter-toms, line 679: possible div by 0 here ?

instaclustr/resource_cluster.go
```diff
@@ -710,6 +732,15 @@ func resourceClusterRead(d *schema.ResourceData, meta interface{}) error{
 	return nil
 }
 
+func stringInSlice(a string, list []string) bool {
```
peter-toms, line 735: [desired] this function could go in the utils file (as dirty as that is, its in keeping with what we have)

acc_test/resource_cluster_test.go
```diff
@@ -231,6 +233,33 @@ func TestAccClusterCustomVPC(t *testing.T) {
 	})
 }
 
+func TestAccKafkaZookeeperContactPoints(t *testing.T) {
```
peter-toms, line 236: [required] lets see a unit test with an odd number of AZs for good measure

peter-toms (changes requested):

Makefile
Chrismelba-Instaclustr: Yeah, I misread the docs. Have made it look more instaclustr-y, though it still just fails to resolve then points at a local folder which is super weird imo

instaclustr/resource_cluster.go
Chrismelba-Instaclustr: Don't think that should be possible. As long as the api gives a rack associated with any node it'll be >0. (also, I just auto-changed spacing, so not really *my* code)

instaclustr/resource_cluster.go
Chrismelba-Instaclustr: it amazes me how many useful functions go doesn't seem to have. It doesn't have a built in max function!

Agreed though, will do that.

acc_test/resource_cluster_test.go
Chrismelba-Instaclustr: built a second test on Cassandra with 3 racks 1 node per rack. Running it now.

instaclustr/resource_cluster.go
peter-toms: think we should guard against that (im imagining a scenario where the nodes are being provisioned for example) , a simple if rackCount >0 otherwise nodesPerRack = 0 should sort it ?

instaclustr/resource_cluster.go
peter-toms: yup its very low coverage from this side of things

PR #99, opened 2021-05-04
Move ForceNew for bundle options to make it take effect

instajonno:
This fixes the issue we were seeing with clusters not being recreated when they should, however it may contain a gotcha in that if you don't specify values and allow them to be default, the first plan/apply creates the cluster but subsequent ones will try and recreate it due to our API returning a value but the plan saying null, for example this from `terraform out`:

    ~ bundle {
        ~ options = { # forces replacement
            - "auto_create_topics"   = "true" -> null
            - "client_encryption"    = "false" -> null
            - "delete_topics"        = "true" -> null
            - "number_partitions"    = "3" -> null
            - "zookeeper_node_count" = "3" -> null
          }

I can't think of an elegant way around that, given it is correct in a strict interpretation of `ForceNew`. 

I also suspect the optional flags are redundant in the schema map, but I haven't tested removing them. In the worst case they do nothing anyway.


instajonno:
Closing this PR after internal discussions around side effects from this change. My recommendation was to approach this in a broader manner to avoid unnecessary recreation of resources.

PR #113, opened 2021-05-25
Ins 11057 multi dc provisioning bundles option

wjjmjh:


evanderploeg (approved):

acc_test/data/valid_basic_multi_DC_provisioning.tf
```diff
@@ -0,0 +1,50 @@
...
+    options = {
+      auth_n_authz = true
+      use_private_broadcast_rpc_address = false
+      client_encryption = false
+      lucene_enabled = false
+      continuous_backup_enabled = true
+    }
+  }
+
+  data_centres {
```
willmasseyic, line 23: [question] don't these `data_centres` need bundles?

willmasseyic: I don't think this file is used anywhere?

instaclustr/resource_cluster.go
```diff
@@ -438,6 +610,14 @@ func resourceClusterCreate(d *schema.ResourceData, meta interface{}) error {
 		createData.RackAllocation = &rackAllocation
 	}
 
+	// for multi-DC cluster
+	if len(dataCentres) > 1 {
+		createData.RackAllocation = createData.DataCentres[0].RackAllocation
```
willmasseyic, line 615: [question] can we rely on the ordering here? The `DataCentres` collection is `schema.TypeSet`, could we get ourselves into a strange situation here?

> TypeSet implements set behavior and is used to represent an unordered collection of items, meaning that their ordering specified does not need to be consistent, and the ordering itself has no impact on the behavior of the resource.
> 
> The elements of a set can be any of the other types allowed by Terraform, including another schema. Set items cannot be repeated.

https://www.terraform.io/docs/extend/schemas/schema-types.html#typeset

instaclustr/resource_cluster.go
```diff
@@ -637,66 +816,82 @@ func resourceClusterRead(d *schema.ResourceData, meta interface{}) error {
...
+			return fmt.Errorf("[Error] Error reading cluster, rack allocation could not be derived: %s", err)
+		}
+
+		if len(cluster.DataCentres[0].ResizeTargetNodeSize) > 0 {
+			nodeSize = cluster.DataCentres[0].ResizeTargetNodeSize
+		}
+
+		d.Set("node_size", nodeSize)
+		d.Set("data_centre", cluster.DataCentres[0].Name)
+		d.Set("cluster_network", cluster.DataCentres[0].CdcNetwork)
```
willmasseyic, line 874: [question] we seem to be referring to DataCentres[0] here but isn't this for a single DC cluster which wont have any `DataCentres` only `DataCentre`? Same with line 851 above

instaclustr/resource_cluster.go
```diff
@@ -637,66 +816,82 @@ func resourceClusterRead(d *schema.ResourceData, meta interface{}) error {
...
-	d.Set("cluster_provider", processedClusterProvider)
-
-	bundles, err := getBundlesFromCluster(cluster)
-	if err != nil {
-		return err
-	}
-
-	if err := d.Set("bundle", bundles); err != nil {
-		return fmt.Errorf("[Error] Error reading cluster: %s", err)
+	if isClusterSingleDataCentre(*cluster) {
```
willmasseyic, line 819: [question] can this be moved to the conditional below?

acc_test/data/valid_basic_multi_DC_provisioning.tf
wjjmjh: removed & resolved.

willmasseyic (approved):

instaclustr/resource_cluster.go
wjjmjh: good point! & resolved.

evanderploeg (approved):

PR #118, opened 2021-06-25
Add Redis Password Auth Support

WeiXinFam-Insta:
Modify cluster resource to include the Redis bundle password auth option
Internal Reference: INS-14566

### DIFF
--- a/instaclustr/resource_cluster.go
+++ b/instaclustr/resource_cluster.go
@@ -520,6 +520,11 @@ func resourceCluster() *schema.Resource {
 										Optional: true,
 										ForceNew: true,
 									},
+									"redis_password_auth": {
+										Type:     schema.TypeBool,
+										Optional: true,
+										ForceNew: true,
+									},
 								},
 							},
 						},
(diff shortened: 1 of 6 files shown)


instaclustr/resource_cluster.go
```diff
@@ -520,6 +520,11 @@ func resourceCluster() *schema.Resource {
 										Optional: true,
 										ForceNew: true,
 									},
+									"redis_password_auth": {
```
charlese-instaclustr, line 523: Might be worth adding this parameter to the tested Redis config in valid_redis_cluster_create.tf , so that we have a test case covering the change?

WeiXinFam-Insta: Reference to INS-14580

instaclustr/resource_cluster.go
```diff
@@ -520,6 +520,11 @@ func resourceCluster() *schema.Resource {
...
+										Type:     schema.TypeBool,
+										Optional: true,
+										ForceNew: true,
```
Chrismelba-Instaclustr, line 526: Does changing this have to force a new resource? Can we add password auth after creating the cluster?

WeiXinFam-Insta: We normally provision a new cluster if we want the cluster to be client encrypted

Chrismelba-Instaclustr (approved):

charlese-instaclustr (approved):
LGTM

PR #121, opened 2021-07-07
Changing sasl_scram_mechanism to a top-level field for the Kafka User Resource

charlese-instaclustr:
Moving the sasl_scram_mechanism from a nested 'options' map to being a top level field for the Kafka user resource


peter-toms (approved):

zegelin (approved):
LGTM

PR #122, opened 2021-08-08
add clusters data source

mootpt:
This is based on: https://developer.instaclustr.com/#operation/listAllUserClusters

Seems like there were some inconsistencies between the request body schema [here]( https://developer.instaclustr.com/#operation/extendedProvisionRequestHandler) and the response body [here]( https://developer.instaclustr.com/#operation/listAllUserClusters).

A future implementation may instead retrieve all active cluster ids and use [this](https://developer.instaclustr.com/#operation/clusterDetailsTerraformHandler). Though it would be better if the API response for List Active Cluster was consistent with the [cluster status for terraform imports](https://developer.instaclustr.com/#operation/clusterDetailsTerraformHandler).

None the less, this PR works for our use case as we just need a way to iterate over all the clusterIds for a few resources.


Chrismelba-Instaclustr:
Thanks @mootpt for this contribution. I agree that this PR could be made even more useful if we made some back-end changes to our API return more information.

If you're ok to wait, we will take this into the instaclustr backlog to prioritise backend API updates that will allow us to expand the intent of this ticket.

I know you're in contact with support@instaclustr.com, so feel free to reach out to either them or here as we work through this feature.

arjunrajinstaclustr:
Yes that sounds good, we shall begin reviewing this along to bring this change in now, while trying to keep forward compatibility in mind.

instaclustr/data_source_clusters.go
```diff
@@ -0,0 +1,85 @@
...
+	log.Print("[INFO] Listing all active clusters")
+	clusterList, err := client.ListClusters()
+	if err != nil {
+		return fmt.Errorf("[Error] Error reading cluster credentials: #{err}")
+	}
+
+	clusters := make([]map[string]interface{}, 0)
+
+	for _, c := range *clusterList {
+		cluster := make(map[string]interface{})
```
arjunrajinstaclustr, line 72: Please ignore if too hard - but could we use `mapstructure` to map the response model to the schema?

```
mapstructure.Decode(c, &cluster)
```

might do the trick, provided that the `ClusterListItem` has the required mapstructure metadata (see `ClusterProvider` for example)

instaclustr/data_source_clusters.go
```diff
@@ -0,0 +1,85 @@
...
+					Schema: map[string]*schema.Schema{
+						"cluster_id": {
+							Type:     schema.TypeString,
+							Computed: true,
+						},
+						"cluster_name": {
+							Type:     schema.TypeString,
+							Computed: true,
+						},
+						"cassandra_version": {
```
arjunrajinstaclustr, line 28: Do you have a current need for this field? Else for forward compatibility purposes for when we decide to properly hook in the right APIs and schema to this data source, could we remove `cassandra_version` ?

arjunrajinstaclustr (changes requested):
Thanks for the awesome PR! Left some minor comments in.

There's also a version bump in the makefile that needs to be done in this change before we can update the provider.

Just some heads up - since we usually add acceptance tests for new features, but the acceptance tests have a cost associated with them, we will shortly make a pull request to your repository with the acceptance test added so we can get the test in before merging this change.

instaclustr/data_source_clusters.go
```diff
@@ -0,0 +1,85 @@
+package instaclustr
+
+import (
+	"fmt"
+	"log"
+
+	"github.com/hashicorp/terraform/helper/schema"
+)
+
+func dataSourceClustersList() *schema.Resource {
```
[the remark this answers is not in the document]
arjunrajinstaclustr: Added the docs

instaclustr/data_source_clusters.go
arjunrajinstaclustr: Removed this field

arjunrajinstaclustr (approved):

michael-carter-instaclustr (approved):

PR #126, opened 2021-09-28
Kafka user override

h3nd24:
Internal Reference: INS-16007

To match the API capability for Kafka user creation: override-existing-user option. Removed the checking if user already exists in the Terraform Provider, the ramification is we can't add pre-existing user, e.g., ickafka to the Terraform Provider.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -3,7 +3,7 @@ BIN_NAME="terraform-provider-instaclustr"
 
 
 # for VERSION, don't add prefix "v", e.g., use "1.9.8" instead of "v1.9.8" as it could break circleCI stuff
-VERSION=1.12.1
+VERSION=1.13.1
 INSTALL_FOLDER=$(HOME)/.terraform.d/plugins/terraform.instaclustr.com/instaclustr/instaclustr/$(VERSION)/darwin_amd64
 
 
--- a/Makefile
+++ b/Makefile
@@ -3,7 +3,7 @@ BIN_NAME="terraform-provider-instaclustr"
 
 
 # for VERSION, don't add prefix "v", e.g., use "1.9.8" instead of "v1.9.8" as it could break circleCI stuff
-VERSION=1.13.1
+VERSION=1.13.0
 INSTALL_FOLDER=$(HOME)/.terraform.d/plugins/terraform.instaclustr.com/instaclustr/instaclustr/$(VERSION)/darwin_amd64
 
 
--- a/acc_test/data/invalid_kafka_user_create_duplicate.tf
+++ b/acc_test/data/invalid_kafka_user_create_duplicate.tf
@@ -1,4 +1,4 @@
-// This is part of testing "kafka user" suite, 4 of 5
+// This is part of testing "kafka user" suite, 5 of 5 
 provider "instaclustr" {
   username = "%s"
   api_key = "%s"
@@ -35,10 +35,25 @@ resource "instaclustr_cluster" "kafka_cluster" {
   }
 }
 
-resource "instaclustr_kafka_user" "kafka_user_charlie_invalid" {
+resource "instaclustr_kafka_user" "kafka_user_charlie" {
+  cluster_id = "${instaclustr_cluster.kafka_cluster.id}"
+  username = "%s"
+  password = "%s"
+  initial_permissions = "none"
+}
+
+resource "instaclustr_kafka_user" "kafka_user_charlie_scram-sha-512" {
   cluster_id          = "${instaclustr_cluster.kafka_cluster.id}"
   username            = "%s"
   password            = "%s"
   initial_permissions = "none"
-  authentication_mechanism = "ExpectedToFail"
+  authentication_mechanism = "SCRAM-SHA-512"
+}
+
+resource "instaclustr_kafka_user" "kafka_user_charlie_double" {
+  cluster_id = "${instaclustr_cluster.kafka_cluster.id}"
+  username = "%s"
+  password = "%s"
+  initial_permissions = "none"
+  override_existing_user = false
 }
--- a/acc_test/data/invalid_kafka_user_create_duplicate.tf
+++ b/acc_test/data/invalid_kafka_user_create_duplicate.tf
@@ -7,7 +7,7 @@ provider "instaclustr" {
 
 resource "instaclustr_cluster" "kafka_cluster" {
   cluster_name = "example_kafka_tf_test"
-  node_size = "KFK-DEV-t4g.medium-80"
+  node_size = "%s"
   data_centre = "US_WEST_2"
   sla_tier = "NON_PRODUCTION"
   cluster_network = "192.168.0.0/18"
@@ -22,7 +22,7 @@ resource "instaclustr_cluster" "kafka_cluster" {
 
   bundle {
     bundle = "KAFKA"
-    version = "apache-kafka:2.7.1"
+    version = "%s"
     options = {
       auto_create_topics = true
       client_encryption = false
@@ -33,7 +33,7 @@ resource "instaclustr_cluster" "kafka_cluster" {
       zookeeper_node_count = 3
     }
   }
-}
+} 
 
 resource "instaclustr_kafka_user" "kafka_user_charlie" {
   cluster_id = "${instaclustr_cluster.kafka_cluster.id}"
@@ -57,3 +57,4 @@ resource "instaclustr_kafka_user" "kafka_user_charlie_double" {
   initial_permissions = "none"
   override_existing_user = false
 }
+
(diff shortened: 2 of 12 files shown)


Makefile
```diff
@@ -3,7 +3,7 @@ BIN_NAME="terraform-provider-instaclustr"
 
 
 # for VERSION, don't add prefix "v", e.g., use "1.9.8" instead of "v1.9.8" as it could break circleCI stuff
-VERSION=1.12.1
+VERSION=1.13.1
```
hail-insta, line 6: should be `1.31.0`?

acc_test/data/invalid_kafka_user_create_duplicate.tf
```diff
@@ -0,0 +1,60 @@
+// This is part of testing "kafka user" suite, 5 of 5
```
hail-insta, line 1: adding a new config for testing will add a fair chunk of time to our acceptance tests (considering that this is kafka user test, meaning that it requires RUNNING cluster). I suggest plugging in testing the duplicate user creation in one of the other tests

acc_test/data/invalid_kafka_user_create_duplicate.tf
```diff
@@ -0,0 +1,60 @@
...
+  node_size = "%s"
+  data_centre = "US_WEST_2"
+  sla_tier = "NON_PRODUCTION"
+  cluster_network = "192.168.0.0/18"
+  wait_for_state = "RUNNING"
+  cluster_provider = {
+    name = "AWS_VPC"
+  }
+  rack_allocation = {
+    number_of_racks = 3
```
hail-insta, line 19: if my above comment (consolidating tests) cant be addressed, can the number of racks reduced to mitigate the addition of test duration?

acc_test/data/invalid_kafka_user_create_duplicate.tf
```diff
@@ -0,0 +1,60 @@
...
+    bundle = "KAFKA"
+    version = "%s"
+    options = {
+      auto_create_topics = true
+      client_encryption = false
+      dedicated_zookeeper = true
+      delete_topics = true
+      number_partitions = 3
+      zookeeper_node_size = "%s"
+      zookeeper_node_count = 3
```
hail-insta, line 33: same as above
> if my above comment (consolidating tests) cant be addressed, can the number of racks reduced to mitigate the addition of test duration?

acc_test/data/valid_kafka.tf
```diff
@@ -22,7 +22,7 @@ resource "instaclustr_cluster" "valid" {
 
     bundle {
         bundle = "KAFKA"
-        version = "apache-kafka:2.7.1"
+        version = "apache-kafka:2.7.1.ic1"
```
hail-insta, line 25: can we do similar changes to above (i.e. `%s`? i think thats a good change)

acc_test/resource_kafka_user_test.go
```diff
@@ -30,19 +32,25 @@ func TestKafkaUserResource(t *testing.T) {
 	kafkaUsername3 := "charlie3"
 	oldPassword := "charlie123!"
 	newPassword := "charlie123standard!"
-	zookeeperNodeSize := "zk-developer-t3.small-20"
+	kafkaNodeSize := "KFK-DEV-t4g.medium-80"
+	zookeeperNodeSize := "KDZ-DEV-t4g.small-30"
+	kafkaVersion := "apache-kafka:2.7.1.ic1"
```
hail-insta, line 37: can we use artefact version? (`2.7.1`) because if we update this, and deprecate `ic1` (which is technically our internal versioning system) it will break the test

acc_test/resource_kafka_user_test.go
```diff
@@ -30,19 +32,25 @@ func TestKafkaUserResource(t *testing.T) {
...
+	createClusterConfig := fmt.Sprintf(string(configBytes1), username, apiKey, hostname, kafkaNodeSize, kafkaVersion, zookeeperNodeSize)
+	createKafkaUserConfig := fmt.Sprintf(string(configBytes2), username, apiKey, hostname, kafkaNodeSize, kafkaVersion, zookeeperNodeSize,
 		kafkaUsername1, oldPassword,
 		kafkaUsername2, oldPassword)
-	createKafkaUserListConfig := fmt.Sprintf(string(configBytes3), username, apiKey, hostname, zookeeperNodeSize,
+	createKafkaUserListConfig := fmt.Sprintf(string(configBytes3), username, apiKey, hostname, kafkaNodeSize, kafkaVersion, zookeeperNodeSize,
 		kafkaUsername1, oldPassword,
 		kafkaUsername2, oldPassword)
-	updateKafkaUserConfig := fmt.Sprintf(string(configBytes3), username, apiKey, hostname, zookeeperNodeSize,
+	updateKafkaUserConfig := fmt.Sprintf(string(configBytes3), username, apiKey, hostname, kafkaNodeSize, kafkaVersion, zookeeperNodeSize,
```
hail-insta, line 46: these are good changes!

instaclustr/resource_cluster.go
```diff
@@ -1160,7 +1160,7 @@ func resourceClusterRead(d *schema.ResourceData, meta interface{}) error {
 	for _, dataCentre := range cluster.DataCentres {
 		for _, node := range dataCentre.Nodes {
 			if !stringInSlice(node.Rack, azList) {
-				if !strings.HasPrefix(node.Size, "zk-") {
+				if !strings.HasPrefix(node.Size, "zk-") && !strings.HasPrefix(node.Size, "KDZ-") {
```
hail-insta, line 1163: i know these were already there, but there are now 3 duplicate and would be good to put it in a single place (and possibly comment why its there)

acc_test/resource_kafka_user_test.go
```diff
@@ -77,6 +85,10 @@ func TestKafkaUserResource(t *testing.T) {
 				Config: updateKafkaUserConfig,
 				Check:  checkKafkaUserUpdated(newPassword),
 			},
+			{
+				Config: invalidKafkaUserCreateConfigDuplicate,
+				ExpectError: regexp.MustCompile("A Kafka user with this username already exists on this cluster."),
```
hail-insta, line 90: question - is this a desired format of error message? in this situation, terraform provider will just return the API errer response, however it seems like it used to be handled on the provider side as well in the code below

```go
// just use linear search for now to check if the user going to be created is already in the system
	usernameList, err := client.ReadKafkaUserList(cluster_id)
	if err != nil {
		return fmt.Errorf("[Error] Error retrieving kafka user list: %s", err)
	}
	for _, str := range usernameList {
		if str == username {
			// user is already set, so we don't change anything
			d.SetId(fmt.Sprintf("%s-%s", cluster_id, username))
			log.Printf("[INFO] Kafka user %s already exists in %s.", username, cluster_id)
			return nil
		}
	}
```

wondering if that should be modified to fail with graceful error message crafted in the terraform provider instead of removing it?

instaclustr/resource_kafka_user.go
```diff
@@ -70,22 +76,9 @@ func resourceKafkaUserCreate(d *schema.ResourceData, meta interface{}) error {
 		return fmt.Errorf("[Error] Cluster %s is not RUNNING. Currently in %s state", cluster_id, cluster.ClusterStatus)
 	}
 
-	// just use linear search for now to check if the user going to be created is already in the system
-	usernameList, err := client.ReadKafkaUserList(cluster_id)
```
hail-insta, line 74: see my comment above

hail-insta (changes requested):

acc_test/resource_kafka_user_test.go
h3nd24: the problem is the asymmetry with the return value thing, if I give 2.7.1 the plan will not be empty because the API returns apache-kafka:2.7.1.ic1

acc_test/resource_kafka_user_test.go
h3nd24: So we removed the handling in terraform provider in favour of the handling in the API side. The API returns a proper error instead of 500, so the error should propagate

instaclustr/resource_cluster.go
```diff
@@ -1409,3 +1408,10 @@ func isClusterSingleDataCentre(cluster Cluster) bool {
 	}
 	return false
 }
+
+// Currently, there is no API to tell if a node should be included as the main contact point
+// or calculated in the rack allocation scheme (that is not returned by the API).
+// Dedicated ZooKeeper nodes fall into this category and require a specific handling by the provider
+func isDedicatedZookeeperNodeSize(nodeSize String) bool {
+	return strings.HasPrefix(nodeSize, "zk-") || !strings.HasPrefix(nodeSize, "KDZ-")
```
hail-insta: is the second condition meant to be `!`?

hail-insta (approved):

hail-insta (approved):

amrutha-shanbhag (approved):

PR #128, opened 2021-10-05
Fix Redis terraform fields

WeiXinFam-Insta:
- Redis would throw an error if the plan consists of `rack_allocation`
- Redis Bundle Options would trigger a `forceNew` operation


instaclustr/resource_cluster.go
```diff
@@ -565,6 +567,29 @@ func resourceCluster() *schema.Resource {
...
+	if _, isBundle := diff.GetOk("bundle"); isBundle {
+		bundle := diff.Get("bundle").([]interface{})
+		bundleMap := bundle[0].(map[string] interface{})
+
+		// Mainly check Single DC Redis Cluster
+		if bundleMap["bundle"] == "REDIS" {
+			if _, isRackAllocationSet := diff.GetOk("rack_allocation"); isRackAllocationSet {
+				return fmt.Errorf("[Error] 'rack_allocation' is not supported in REDIS")
+			}
+			// Remove this logic once INS-13970 is implemented
```
Mitchel-Stig-Instaclustr, line 581: might want to add a comment to that ticket to make sure it gets removed?

PR #135, opened 2021-10-29
Ins 14478 Refreshing terraform state from single to multi DC cluster fail

yuzhouchen-instaclustr:


instaclustr/resource_cluster.go
```diff
@@ -1235,6 +1240,20 @@ func resourceClusterRead(d *schema.ResourceData, meta interface{}) error {
 	return nil
 }
 
+func deleteAttributesConflictWithDataCentres(d *schema.ResourceData) error {
```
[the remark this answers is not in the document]
yuzhouchen-instaclustr: In scenario 1 terraform won't run the plan as it is not legal.
In scenario 2, IIRC terraform will try to destroy the original DC and create the new ones (which is not ideal, but it is not what this fix is trying to fix)

acc_test/test_utils.go
```diff
@@ -54,3 +56,26 @@ func checkClusterRunning(resourceName, hostname, username, apiKey string) resour
...
+		body, _ := ioutil.ReadFile(requestBody)
+
+		// Adding a new DC to an existing cluster
+		url := fmt.Sprintf("%s/provisioning/v1/%s/cluster-data-centres", hostname, id)
+		resp, err := client.MakeRequest(url, "POST", body)
+
+		if err != nil {
+			return err
+		}
+		bodyText, err := ioutil.ReadAll(resp.Body)
```
chrisw-instaclustr, line 75: Should we return an error in the case of a failure to read the response body here?

instaclustr/resource_cluster.go
```diff
@@ -1235,6 +1240,20 @@ func resourceClusterRead(d *schema.ResourceData, meta interface{}) error {
...
+	clusterResource := resourceCluster()
```
chrisw-instaclustr, line 1244: How nasty would it be to mock out `resourceCluster()` here so that we can stick some unit tests around this?

instaclustr/resource_cluster.go
```diff
@@ -1235,6 +1240,20 @@ func resourceClusterRead(d *schema.ResourceData, meta interface{}) error {
...
+	for key, value := range clusterResource.Schema {
+		for _, conflictsWith := range value.ConflictsWith {
+			if _, exist := d.GetOk(key); exist && conflictsWith == "data_centres" {
+				if err := d.Set(key, value.Type.Zero()); err != nil {
+					return err
```
chrisw-instaclustr, line 1249: How perspicuous is this error?  If something does go wrong, will folks easily be able to track down the source of the error?

acc_test/test_utils.go
yuzhouchen-instaclustr: Yep added return err below

instaclustr/resource_cluster.go
yuzhouchen-instaclustr: I don't believe d.Set here should have an error ever, but if it does the error should be pretty explanatory

instaclustr/resource_cluster.go
yuzhouchen-instaclustr: A unit test has been added for this function

instaclustr/resource_cluster_unit_test.go
```diff
@@ -550,6 +550,51 @@ func TestCreateVpcPeeringRequestLegacy(t *testing.T) {
...
+		"attributeC": "C",
+	}
+
+	d := schema.TestResourceDataRaw(t, clusterSchema, resourceDataMap)
+
+	if err := deleteAttributesConflictWithDataCentres(clusterSchema, d); err != nil {
+		t.Fatalf("Unexpected error occured during deletion %s", err)
+	}
+
+	helper := func(attribute string, expected interface{}) {
```
chrisw-instaclustr, line 587: Can we please use a more descriptive name than "helper"?  How about something like `checkAttributeValue`?

chrisw-instaclustr (approved):

arjunrajinstaclustr (approved):

instaclustr/resource_cluster.go
```diff
@@ -1176,7 +1176,12 @@ func resourceClusterRead(d *schema.ResourceData, meta interface{}) error {
 		}
 		d.Set("data_centre", cluster.DataCentres[0].Name)
 		d.Set("cluster_network", cluster.DataCentres[0].CdcNetwork)
+
+		if err := deleteAttributesConflict(resourceCluster().Schema, d, "data_centre"); err != nil {
```
yuzhouchen-instaclustr, line 1180: We need to delete attributes conflicting with `data_centre` if for whatever reason the actual infrastructure is single-DC and the terraform state is multi-DC

chrisw-instaclustr (approved):

arjunrajinstaclustr (dismissed):

chrisw-instaclustr (dismissed):

arjunrajinstaclustr (approved):

evanderploeg (dismissed):

arjunrajinstaclustr (approved):

evanderploeg (approved):

PR #137, opened 2021-11-12
INS-16006 Terraform Provider for Kafka Topics Management

YuWang24:
Add kafka topic management.


acc_test/resource_kafka_topic_test.go
```diff
@@ -0,0 +1,211 @@
...
+		topic2)
+	invalidKafkaTopicCreateConfigDuplicate := fmt.Sprintf(string(configBytes4), username, apiKey, hostname, kafkaNodeSize, kafkaVersion, zookeeperNodeSize,
+		topic1,
+		topic2,
+		topic1)
+	updateKafkaTopicConfig := fmt.Sprintf(string(configBytes5), username, apiKey, hostname, kafkaNodeSize, kafkaVersion, zookeeperNodeSize,
+		topic1,
+		topic2)
+
+	resource.Test(t, resource.TestCase{
```
michael-carter-instaclustr, line 51: The acc_tests are really slow and flakey unfortunately (particularly tests that wait for clusters to reach running state). So I'm keen to make sure if we add new tests that they're the absolutely most efficient versions of the tests that can be. Is there any way to minimise the number of clusters we have to wait to reach running?

instaclustr/resource_kafka_topic.go
```diff
@@ -0,0 +1,382 @@
...
+
+// There are two use cases of resourceKafkaTopicUpdate:
+// 1. Providing all the possible configs in resource, and modify the ones you want to update. Then keeping all the
+// configs in the resource.
+// 2. Providing some configs (not all) you want to update, and then keep only these configs in the resource.
+// Terraform always detects that there is an update in case 2 if you run the plan again, because resourceKafkaTopicRead
+// reads all the configs, but there are only some of them provided in resource. For the missing configs, Terraform just
+// sets them as their default values.
+//
+// Although Terraform detects an update, func getChangedConfigMap is used to handle case 2. It makes sure that there
```
michael-carter-instaclustr, line 255: Makes sense. Does Terraform itself have any standard way of dealing with this situation? (e.g. diff suppression or something like that)

instaclustr/resource_kafka_topic.go
```diff
@@ -0,0 +1,382 @@
...
+		err = resourceKafkaTopicRead(d, meta)
+		if err != nil {
+			return err
+		}
+		return nil
+	}
+
+	var kafkaTopicConfigOptions KafkaTopicConfigOptions
+	err = mapstructure.Decode(changedConfigMap, &kafkaTopicConfigOptions)
+	if err != nil {
```
michael-carter-instaclustr, line 290: I read somewhere or other that the preferred way of propagating errors in go is to chain them with context. I'm not a go expert so I'm not sure if that true or not but here is an example anyway:

<pre>
func (thing *Thingo) Thingo() error {
    if err := thing.DoSomething(); err != nil {
        return fmt.Errorf("could not do the thing: %w", err)
    }
    ...
}
</pre>

instaclustr/data_source_kafka_topic_list.go
```diff
@@ -0,0 +1,44 @@
...
+					Type: schema.TypeString,
+				},
+				Computed: true,
+			},
+		},
+	}
+}
+
+func dataSourceKafkaTopicListRead(d *schema.ResourceData, meta interface{}) error {
+	client := meta.(*Config).Client
```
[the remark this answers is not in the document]
YuWang24: I think it's the standard way that how we initiate the api client. Other places also use it in this way. My understanding is that this retrieves info from the provider block:
provider "instaclustr" {
  username = "%s"
  api_key = "%s"
  api_hostname = "%s"
}

Makefile
```diff
@@ -3,7 +3,7 @@ BIN_NAME=terraform-provider-instaclustr
 
 
 # for VERSION, don't add prefix "v", e.g., use "1.9.8" instead of "v1.9.8" as it could break circleCI stuff
-VERSION=1.14.3
+VERSION=1.14.5
```
arjunrajinstaclustr, line 6: Why is this being bumped up by 2?

instaclustr/resource_kafka_topic.go
YuWang24: I played around with it, and did some researches. I changed the implementations several times. But it seems the Terraform way couldn't handle these cases very well.

acc_test/resource_kafka_topic_test.go
YuWang24: For these tests, I am afraid that they need to have the cluster in RUNNING, because the API calls need cluster to be in RUNNING state.

instaclustr/api_client_kafka.go
```diff
@@ -0,0 +1,135 @@
+package instaclustr
+
+import (
+	"encoding/json"
+	"errors"
+	"fmt"
+	"io/ioutil"
+)
+
+func (c *APIClient) ReadKafkaUserList(clusterID string) ([]string, error) {
```
arjunrajinstaclustr, line 10: Thanks for splitting these out!

instaclustr/data_source_kafka_topic_list.go
```diff
@@ -0,0 +1,44 @@
+package instaclustr
```
arjunrajinstaclustr, line 1: There should be examples for the new stuff under examples/main.tf

There also needs to be documentation created under the docs/ folder, which is used by the Terraform Registry.

acc_test/resource_kafka_topic_test.go
michael-carter-instaclustr: So is this one cluster or five?

instaclustr/data_source_kafka_topic_list.go
```diff
@@ -0,0 +1,44 @@
...
+	return &schema.Resource{
+		Read: dataSourceKafkaTopicListRead,
+
+		Schema: map[string]*schema.Schema{
+			"cluster_id": {
+				Type:     schema.TypeString,
+				Required: true,
+			},
+
+			"topics": &schema.Schema{
```
arjunrajinstaclustr, line 20: Just wondering - why is the topic list here limited to a list of strings? Does the API not return the other information we feed into the topic resource?

instaclustr/structs_kafka.go
```diff
@@ -0,0 +1,75 @@
+package instaclustr
```
arjunrajinstaclustr, line 1: Thanks for splitting these out too :)

instaclustr/resource_kafka_topic.go
```diff
@@ -0,0 +1,382 @@
...
+	log.Printf("[INFO] Successfully fetch the config of topic %s.", topic)
+	return nil
+}
+
+// There are two use cases of resourceKafkaTopicUpdate:
+// 1. Providing all the possible configs in resource, and modify the ones you want to update. Then keeping all the
+// configs in the resource.
+// 2. Providing some configs (not all) you want to update, and then keep only these configs in the resource.
+// Terraform always detects that there is an update in case 2 if you run the plan again, because resourceKafkaTopicRead
+// reads all the configs, but there are only some of them provided in resource. For the missing configs, Terraform just
```
arjunrajinstaclustr, line 252: If the default values line up with the API's defaults, shouldn't terraform detect no change?

instaclustr/resource_kafka_topic.go
```diff
@@ -0,0 +1,382 @@
...
+// will not be a real update if the configs in the resources are not changed. So, it's safe to keep only partial of
+// configs in the config{} block, there won't be an update request if their values are same as what have been read.
+func resourceKafkaTopicUpdate(d *schema.ResourceData, meta interface{}) error {
+	cluster_id := d.Get("cluster_id").(string)
+	topic := d.Get("topic").(string)
+	log.Printf("[INFO] Updating the config of Kafka topic %s in %s.", topic, cluster_id)
+
+	// This is for when only replication_factor or partitions are changed, but we don't support changing them currently
+	if !d.HasChange("config") {
+		log.Printf("[INFO] Currently we only supprot updating topic's config. There are no changes in topic %s's config.", topic)
```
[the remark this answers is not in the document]
YuWang24: Yeah, will modify this.

instaclustr/resource_kafka_topic.go
YuWang24: Yeah, you're right, will add the context.

acc_test/resource_kafka_topic_test.go
YuWang24: Only one cluster, one cluster for all the tests.

instaclustr/resource_kafka_topic.go
```diff
@@ -0,0 +1,382 @@
...
+}
+
+func resourceKafkaTopicDelete(d *schema.ResourceData, meta interface{}) error {
+	cluster_id := d.Get("cluster_id").(string)
+	topic := d.Get("topic").(string)
+	log.Printf("[INFO] Deleting Kafka topic %s in %s.", topic, cluster_id)
+	client := meta.(*Config).Client
+
+	err := client.DeleteKafkaTopic(cluster_id, topic)
+	if err != nil {
```
arjunrajinstaclustr, line 363: If a Terraform created Kafka topic is deleted outside of Terraform, is deleting the topic through Terraform handled gracefully?

arjunrajinstaclustr (changes requested):

instaclustr/data_source_kafka_topic_list.go
YuWang24: Yeah, we have another ticket for that because this topic management feature still needs to wait for the NA rollout to function in production.

Makefile
YuWang24: I saw there is another PR marked as 1.14.4. I think it might be go into preprod before me.

instaclustr/api_client_kafka.go
YuWang24: No worries!

instaclustr/data_source_kafka_topic_list.go
YuWang24: No, for reading topic list, it only returns the list of topics, no other information.

acc_test/resource_kafka_topic_test.go
michael-carter-instaclustr: OK Great. We already have acc tests that create Kafka clusters and wait for them to reach running. Are there opportunities to combine? (To reduce the time we spend waiting and the $$ we spend on creating)

instaclustr/data_source_kafka_topic_list.go
michael-carter-instaclustr: Every release of the terraform provider is public. I don't think we should be splitting out doco tickets into separate things. Sounds like we should wait for the NA rollout before adding this feature into this repo

instaclustr/data_source_kafka_topic_list.go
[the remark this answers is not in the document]
michael-carter-instaclustr: OK, fair enough

instaclustr/resource_kafka_topic.go
YuWang24: There are no API's defaults, it will read the current configs for a topic. And it depends whether there is a full list of topic config provided in the resource. Like I said in the comment, if users only provide a partial list of config in the resource, Terraform will detect other missing configs, and set them as null.

instaclustr/resource_kafka_topic.go
YuWang24: Oh, I haven't tried that, need to have a quick try.

instaclustr/resource_kafka_topic.go
michael-carter-instaclustr: if you are following this method, I believe you should use the %w specifier with 'err' operands

acc_test/resource_kafka_topic_test.go
YuWang24: I think for every acc test, it will destroy the cluster after all steps are completed. Unless I combine resource_kafka_topic_test.go and resource_kafka_user_test.go together. But I don't know whether this is a good idea, because these two tests are testing different things.

michael-carter-instaclustr: From a cleanliness of testing point of view I agree. But in this case where we're doing long running operations that cost real time and money, I think it's worth considering if the cluster is really part of your test, or whether it is simply a platform for running your tests of users and topics on. In which case, Why should we create more platforms than we need to?

instaclustr/resource_kafka_topic.go
YuWang24: Ah, forgot about this, will add it.

acc_test/resource_kafka_topic_test.go
YuWang24: Okay, that sounds reasonable. Then I will combine both users and topics into a more general test file resource_kafka_test.go. Considering we're going to add ACL management soon, I think it worths combining them into one to save time and money as only one cluster is needed for all these operations.

instaclustr/data_source_kafka_topic_list.go
YuWang24: Yeah, talked with @amrutha-shanbhag, she thought we could continue the review process, but hold the release of this PR. This will come with the final release of the topic management.

instaclustr/api_client_kafka_unit_test.go
```diff
@@ -0,0 +1,63 @@
...
+	"testing"
+)
+
+func TestAPIClientCreateKafkaTopic(t *testing.T) {
+	filename := "data/valid_kafka_topic.json"
+	jsonStr, err := ioutil.ReadFile(filename)
+	if err != nil {
+		t.Fatalf("Failed to load %s: %s", filename, err)
+	}
+	client := SetupMock(t, "should-be-uuid/kafka/topics", `{"id":"should-be-uuid"}`, 201)
```
h3nd24, line 15: what is this `should-be-uuid` thing? I suppose I haven't used SetupMock before, so I'm just not sure what this is doing

Makefile
arjunrajinstaclustr: Ah I see. You'll want to sync with releases then since you won't be able to merge your branch if it's not up to date

instaclustr/resource_kafka_topic.go
```diff
@@ -0,0 +1,382 @@
...
+		key := fmt.Sprintf("config.0.%s", k)
+		valueOfV := reflect.ValueOf(decodedConfigMap[k])
+		if d.HasChange(key) {
+			before, after := d.GetChange(key)
+			isBeforeZeroValue := before == reflect.Zero(reflect.TypeOf(before)).Interface()
+			isAfterZeroValue := after == reflect.Zero(reflect.TypeOf(after)).Interface()
+			if !isBeforeZeroValue && !isAfterZeroValue {
+				changedConfigMap[k] = v
+			}
+			if !isBeforeZeroValue && isAfterZeroValue && (valueOfV.Kind() == reflect.Ptr) {
```
h3nd24, line 345: what about keys that can be genuinely 0? like `flush.messages`?

instaclustr/resource_kafka_topic.go
```diff
@@ -0,0 +1,382 @@
...
+	err = mapstructure.Decode(kafkaTopicConfig, &decodedConfigMap)
+	if err != nil {
+		return nil, err
+	}
+
+	// The following block is to get what configs are indeed changed
+	for k, v := range newConfig {
+		key := fmt.Sprintf("config.0.%s", k)
+		valueOfV := reflect.ValueOf(decodedConfigMap[k])
+		if d.HasChange(key) {
```
h3nd24, line 338: I haven't played with this, but what happens if you just include the keys that hasChange? that is, without the additional processing using reflection

acc_test/data/kafka_topic_create_topic.tf
```diff
@@ -0,0 +1,50 @@
...
+
+resource "instaclustr_kafka_topic" "kafka_topic_test" {
+  cluster_id = "${instaclustr_cluster.kafka_cluster.id}"
+  topic = "%s"
+  replication_factor = 3
+  partitions = 3
+}
+
+resource "instaclustr_kafka_topic" "kafka_topic_test2" {
+  cluster_id = "${instaclustr_cluster.kafka_cluster.id}"
```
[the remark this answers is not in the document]
YuWang24: Mainly used to test the list topic endpoints. To make sure we have the both created topics in the topic list.

instaclustr/api_client_kafka_unit_test.go
YuWang24: You can imagine should-be-uuid as a cluster-id here.

instaclustr/resource_kafka_topic.go
YuWang24: This is a really good question, I thought a lot about this, that's also the reason why I set some of the configs as * pointers. My thought was to set the configs that might be set as zero-value and configs that have a default zero-value to pointers. So I look at this https://kafka.apache.org/23/documentation.html#topicconfigs, but I found there are many of them. But for configs like flush.messages, which has a very huge default number, I assume it will not be set to 0.

instaclustr/resource_kafka_topic.go
YuWang24: What do you mean? reflection is not used here, it's for later usage.

h3nd24: yeah I mean further processing using reflection and stuff, I think I missed the party a bit, your discussion with Arjun above seems to answered my question

instaclustr/resource_kafka_topic.go
h3nd24: that makes sense to me

instaclustr/resource_kafka_topic.go
```diff
@@ -0,0 +1,398 @@
...
+		return fmt.Errorf("[Error] Error creating kafka topic update request: %w", err)
+	}
+
+	err = client.UpdateKafkaTopic(cluster_id, topic, jsonStr)
+	if err != nil {
+		return fmt.Errorf("[Error] Error updating the config of Kafka topic %s: %w", topic, err)
+	}
+
+	//Reading updated configs
+	err = resourceKafkaTopicRead(d, meta)
```
h3nd24, line 325: Why Is there a need to manually invoke the reading life cycle?

YuWang24: This is basically to keep the terraform.tfstate in sync. Especially when there are only partial configs are provided in the resource. This read after update will not set the missing configs as their zero-values, instead it reads the real values of them.

acc_test/resource_kafka_topic_test.go
YuWang24: Acceptance tests of Kafka user and topic management are combined together.

acc_test/resource_kafka_test.go
```diff
@@ -49,10 +56,25 @@ func TestKafkaUserResource(t *testing.T) {
 	invalidKafkaUserCreateConfigDuplicate := fmt.Sprintf(string(configBytes5), username, apiKey, hostname, kafkaNodeSize, kafkaVersion, zookeeperNodeSize,
 		kafkaUsername1, newPassword,
 		kafkaUsername2, newPassword,
-                kafkaUsername1, oldPassword)
+		kafkaUsername1, oldPassword)
 	invalidKafkaUserCreateConfig := fmt.Sprintf(string(configBytes4), username, apiKey, hostname, kafkaNodeSize, kafkaVersion, zookeeperNodeSize,
 		kafkaUsername3, oldPassword)
 
+	// Kafka topic management tests
+	createKafkaTopicConfig := fmt.Sprintf(string(configBytes6), username, apiKey, hostname, kafkaNodeSize, kafkaVersion, zookeeperNodeSize,
```
michael-carter-instaclustr, line 64: Awesome, thanks for doing this. One question though - do all of the configs now have duplicated resource definitions for the same cluster? (i.e. if we ever update it, we need to update it in five places). If so, is it possible to restructure the data files so that it is only specified once?

YuWang24: Yeah, all the tf files share the same cluster configurations. But the most important configs (the configs that might be changed) are all parameterised. See below:
```
	kafkaUsername1 := "charlie1"
	kafkaUsername2 := "charlie2"
	kafkaUsername3 := "charlie3"
	oldPassword := "charlie123!"
	kafkaNodeSize := "KFK-DEV-t4g.medium-80"
	zookeeperNodeSize := "KDZ-DEV-t4g.small-30"
	kafkaVersion := "apache-kafka:2.7.1.ic1"
	topic1 := "test1"
	topic2 := "test2"
```

docs/resources/kafka_topic.md
```diff
@@ -0,0 +1,151 @@
...
+segment_index_bytes | int | None
+
+<sup>* Indeed an int value, using string because Terraform has difficulties parsing big int, it will lose precision.
+
+## Usage
+There are two use cases for updating a Kafka topic's configs.
+1. Providing all the possible configs in resource, and modify the ones you want to update. Then keeping all the configs in the resource.
+2. Providing some configs (not all) you want to update, and then keep only these configs in the resource.
+
+Terraform always detects that there is an update in case 2 if you run the plan again. Although Terraform detects an update, we make sure
```
arjunrajinstaclustr, line 88: This still isn't great. Terraform says one thing, but we do something else. We might need to re-discuss the option to suppress the diff if the property is null.

arjunrajinstaclustr (changes requested):

docs/resources/kafka_topic.md
```diff
@@ -0,0 +1,151 @@
...
+terraform resources.
+
+### Import a topic and its replication-factor and partitions
+```shell
+terraform import instaclustr_kafka_topic.<resource-name> "<cluster-id>&<topic-name>"
+```
+
+### Import a topic and its configs
+```shell
+terraform import instaclustr_kafka_topic.<resource-name> "<cluster-id>&<topic-name>&config"
```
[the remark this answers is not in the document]
YuWang24: Because there could be cases that our customers only want to manage topics (create/delete) rather than topic configs. Then in this case, I think there is no need to import the topic config.

docs/resources/kafka_topic.md
YuWang24: I don't think there is an easy way to handle this. By saying this in doc, we are actually encouraging our customers to use the first way.  They can store all the configs in the resource and change the ones they want to change. The first way is also better in terms of keeping a record of all the topic configs for latter usages and references.

docs/resources/kafka_topic.md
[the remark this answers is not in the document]
arjunrajinstaclustr: The standard for importing a resource is to have singular behaviour though, not have conditional behaviour based on the input. The input to a terraform config is meant to be an identifier for the resource, not a directive. If the user only cares about the non-config properties, all they need to do is match the config with what is imported, which is a simple thing to do.

instaclustr/resource_kafka_topic.go
```diff
@@ -0,0 +1,419 @@
...
+	if len(idParts) != 2 || idParts[0] == "" || idParts[1] == "" {
+		return nil, fmt.Errorf("[Error] Unexpected format of ID (%q), expected <CLUSTER-ID>&<TOPIC>", d.Id())
+	}
+	d.Set("cluster_id", idParts[0])
+	d.Set("topic", idParts[1])
+	return []*schema.ResourceData{d}, nil
+}
+
+//customDiffSuppressFunc is used to suppress the diff if a config is not provided in the resource.
+func customDiffSuppressFunc(k, old, new string, d *schema.ResourceData) bool {
```
arjunrajinstaclustr, line 417: Should have a separate diff method here based on the type of the property. Because it's possible for "0" to be a legitimate input to a string property.

instaclustr/resource_kafka_topic.go
```diff
@@ -0,0 +1,419 @@
...
+	log.Printf("[INFO] Updating the config of Kafka topic %s in %s.", topic, cluster_id)
+
+	// This is for when only replication_factor or partitions are changed, but we don't support changing them currently
+	if !d.HasChange("config") {
+		log.Printf("[INFO] Currently we only support updating topic's config. There are no changes in topic %s's config.", topic)
+		return nil
+	}
+	client := meta.(*Config).Client
+
+	changedConfigMap, err := getChangedConfigMap(d)
```
arjunrajinstaclustr, line 304: I'm still not sure if this getChangedConfigMap method is necessary. Does the update API require all configuration properties to be provided? What if only the updated properties are provided?

YuWang24: The purpose of this func is to give only the configs that are updated. Rather than giving all the configuration properties.

instaclustr/resource_kafka_topic.go
YuWang24: Yeah, that's true, will add another function for int.

docs/data-sources/kafka_topic.md
```diff
@@ -0,0 +1,27 @@
...
+Property | Description | Default
+---------|-------------|--------
+`cluster_id`|The ID of an existing Instaclustr Kafka managed cluster. |Required
+
+#### Example
+```terraform
+data "instaclustr_kafka_topic_list" "kafka_topic_list" {
+  cluster_id = "${instaclustr_cluster.kafka_cluster.id}"
+}
+```
```
arjunrajinstaclustr, line 26: Could we also have a reference of the properties provided by this data source?

arjunrajinstaclustr (approved):

h3nd24 (approved):

docs/data-sources/kafka_topic.md
h3nd24: OK

PR #139, opened 2021-11-22
Ins 14888 Customer Initiated Resize - Cassandra

yuzhouchen-instaclustr:
Resizable node sizes will be retired, and normal node sizes will become resizable,  need to modify the plugin to reflect the changes in the backend.


acc_test/resource_cluster_test.go
```diff
@@ -165,20 +166,24 @@ func TestAccClusterResize(t *testing.T) {
 				),
 			},
 			{
+				PreConfig: func() {
+					fmt.Println("Sleep for 15 minutes to wait for Cassandra cluster to be ready for resize.")
+					time.Sleep(15 * time.Minute)
```
michael-carter-instaclustr, line 171: Why 15 minutes? That seems a long time to wait. Can't you just wait until it's running?

yuzhouchen-instaclustr: So we will implement a health check in API that will basically stops a Cassandra CDC resize from happening if it's within 15 minutes after it hits running, so we will need to wait for 15 minutes before a resize is possible, I'm not too sure if we should still have this test though since it'll take a long time to run

michael-carter-instaclustr: The time it will take to run is my primary concern, yes. I'd be happier with this sort of test if these tests were parallelised. Unfortunately I don't think they are.

michael-carter-instaclustr: Can you disable the health check in this context at all?

michael-carter-instaclustr: Hmm, shame. Might have to leave it as is for now. Looks like there are some unit test failures on this branch btw

yuzhouchen-instaclustr: yeah we won't get a green run unless the relevant API change is deployed

michael-carter-instaclustr (dismissed):

Mitchel-Stig-Instaclustr (dismissed):

acc_test/resource_cluster_test.go
arjunrajinstaclustr: Why don't we just do a timed retry if the pre-condition fails?

yuzhouchen-instaclustr: it shouldn't make any difference though, it will definitely fail if resizing within 15 minutes after provisioning

hoangvuleInstaclustr (approved):

arjunrajinstaclustr (approved):

hoangvuleInstaclustr (dismissed):

michael-carter-instaclustr (dismissed):

arjunrajinstaclustr (approved):

hoangvuleInstaclustr (dismissed):

PR #142, opened 2021-12-06
INS-13976 Enhance API and Terraform bundle version formats

Jack-Kilrain:
## Change log:
* Created a diff suppression function for bundle version that ignores differences that are not sem-ver changes. For example, all of the below produce no diff:
  * Old (apache-cassandra-3.11.8.ic2) -> New (3.11.8)
  * New (3.11.8)  -> Old (apache-cassandra-3.11.8.ic2)
  * New (3.11.8) -> New (3.11.8)
  * Old (apache-cassandra-3.11.8.ic2) -> Old (apache-cassandra-3.11.8.ic2)
* Added acceptance tests for diff suppression
* Removed mention of legacy version formats


Jack-Kilrain:
FYI acceptance test failure is unrelated to this change, its existing encryption key tests failing. Taking a look into it.

acc_test/resource_cluster_test.go
```diff
@@ -680,3 +680,71 @@ func testCheckResourceStateAttributesDeleted(resourceName string, attributeNames
...
+				Check: resource.ComposeTestCheckFunc(
+					testCheckResourceValid(resourceName),
+				),
+				PlanOnly: true,
+			},
+		},
+	})
+}
+
+func TestCheckNoDiffNewToOldBundleVersionFormat(t *testing.T) {
```
Mitchel-Stig-Instaclustr, line 718: I might be blind but are these tests identical?

acc_test/resource_cluster_test.go
```diff
@@ -680,3 +680,71 @@ func testCheckResourceStateAttributesDeleted(resourceName string, attributeNames
...
+		Steps: []resource.TestStep{
+			{
+				Config: oldVersionConfig,
+				Check: resource.ComposeTestCheckFunc(
+					testCheckResourceValid(resourceName),
+					testCheckResourceCreated(resourceName, hostname, username, apiKey),
+				),
+			},
+			{
+				Config:      newVersionConfig,
```
Mitchel-Stig-Instaclustr, line 708: some weird spacing here

instaclustr/resource_cluster.go
```diff
@@ -584,6 +587,13 @@ func resourceCluster() *schema.Resource {
 		},
 	}
 }
+
+func versionDiffSuppressFunc(k, old string, new string, d *schema.ResourceData) bool {
```
Mitchel-Stig-Instaclustr, line 591: might be nice to have a comment explaining exactly what this is doing?  Im not a regex expert so i can only guess.

Mitchel-Stig-Instaclustr (changes requested):

acc_test/resource_cluster_test.go
EngineersBox: They are different in that one tests there are no plan changes for changing from the legacy version format to the new sem-ver only format, the other tests the inverse transition

acc_test/resource_cluster_test.go
EngineersBox: Not sure why this is here, probably some weird auto formatting differences.

instaclustr/resource_cluster.go
EngineersBox: Yep, can do.

instaclustr/resource_cluster.go
```diff
@@ -23,6 +23,7 @@ var (
 		"RUNNING":     true,
 		"PROVISIONED": true,
 	}
+	semanticVersioningPattern, _ = regexp.Compile("([0-9]*\\.){2}[0-9]")
```
michael-carter-instaclustr, line 26: Looks like you're forcing the last digit to be single only, and allowing '.'s with no numbers.
How about ([0-9]+\\.){2}[0-9]+ instead?

acc_test/resource_cluster_test.go
```diff
@@ -680,3 +680,71 @@ func testCheckResourceStateAttributesDeleted(resourceName string, attributeNames
 		return nil
 	}
 }
+
+func TestCheckNoDiffOldToNewBundleVersionFormat(t *testing.T) {
```
michael-carter-instaclustr, line 684: I'm not keen to add more acc_tests that create more clusters. It slows everything down.
Is there a valid way to unit test this? Or tp piggyback it off a cluster that has already been created as part of these tests?

michael-carter-instaclustr (changes requested):

acc_test/resource_cluster_test.go
EngineersBox: Yeah I should be able to attach to an existing provisioning test. I'll refactor it to that. As for unit testing, I'll take a look into it.

instaclustr/resource_cluster.go
EngineersBox: Ah this is the old regex. I'll change it to the updated version of the regex: [0-9]*\\.[0-9]*(\\.[0-9])?

acc_test/resource_cluster_test.go
```diff
@@ -21,6 +21,8 @@ func AccClusterResourceTestSteps(t *testing.T, testAccProviders map[string]terra
 
 	oriConfig := fmt.Sprintf(string(validConfig), username, apiKey, hostname)
 	updatedConfig := strings.Replace(oriConfig, "testcluster", "newcluster", 1)
+	oldToNewVersionConfig := strings.Replace(updatedConfig, `version = "apache-cassandra-3.11.8.ic2"`, `version = "3.11.8"`, 1)
```
michael-carter-instaclustr, line 24: Hmm, does all our existing test configs use the old format? If so, do you think they should be updated as well?

acc_test/resource_cluster_test.go
```diff
@@ -40,6 +42,34 @@ func AccClusterResourceTestSteps(t *testing.T, testAccProviders map[string]terra
...
 			},
+			{
+				Config: oldToNewVersionConfig,
+				Check: resource.ComposeTestCheckFunc(
+					testCheckResourceValid("valid"),
+				),
+				PlanOnly: true,
+			},
+			{
+				Config: oldToNewVersionConfig,
```
michael-carter-instaclustr, line 53: Why would old to new destroy the cluster? Or am I reading this test step wrong?

instaclustr/resource_cluster.go
```diff
@@ -23,6 +23,7 @@ var (
 		"RUNNING":     true,
 		"PROVISIONED": true,
 	}
+	semanticVersioningPattern, _ = regexp.Compile("[0-9]+\\.[0-9]+(\\.[0-9]+)*")
```
michael-carter-instaclustr, line 26: ok, so this regex allows 2+ versions. so 3.1 is valid, and so is 3.1.2.3.4.5.6.7.7
Is that intentional?
(If so, could be simplified to [0-9]+(\\.[0-9]+)+

acc_test/resource_cluster_test.go
Jack-Kilrain: It needs to be destroyed and re-created in order to change the format of the version string to check the inverse trasition between version formats.

acc_test/resource_cluster_test.go
Jack-Kilrain: Pretty much, that's something I can go through and change too.

instaclustr/resource_cluster.go
EngineersBox: So that last * should be a ?, which will only match duplets or triplets. Since this is always a left-most retrieval any subsequent matched of duplets and triplets are ignored.

michael-carter-instaclustr (approved):

arjunrajinstaclustr (approved):

Mitchel-Stig-Instaclustr (approved):

PR #144, opened 2021-12-13
Kafka Connect Acceptance Test Automation

h3nd24:


acc_test/resource_kafka_test.go
```diff
@@ -41,16 +44,26 @@ func TestKafkaResources(t *testing.T) {
...
 	kafkaUsername1 := "charlie1"
 	kafkaUsername2 := "charlie2"
 	kafkaUsername3 := "charlie3"
 	oldPassword := "charlie123!"
 	newPassword := "charlie123standard!"
 	kafkaNodeSize := "KFK-DEV-t4g.medium-80"
 	zookeeperNodeSize := "KDZ-DEV-t4g.small-30"
-	kafkaVersion := "2.7.1"
+	kafkaVersion := "apache-kafka:3.0.0.ic1"
+	kcVersion := "kafka-connect:3.0.0.ic1"
```
michael-carter-instaclustr, line 58: Use 3.0.0 please, not the internal bundle name

acc_test/resource_kafka_test.go
```diff
@@ -41,16 +44,26 @@ func TestKafkaResources(t *testing.T) {
...
 	topic1 := "test1"
 	topic2 := "test2"
+	azureStorageAccountName := os.Getenv("IC_AZURE_STORAGE_ACCOUNT_NAME")
```
michael-carter-instaclustr, line 61: Are these all already added to CircleCI

acc_test/test_utils.go
```diff
@@ -21,41 +21,55 @@ func getOptionalEnv(key, fallback string) string {
...
-				break
-			}
-			if timePassed > WaitForClusterTimeout {
-				fmt.Printf("\n")
-				return fmt.Errorf("[Error] Timed out waiting for cluster to have the status 'RUNNING'. Current cluster status is '%s'", latestStatus)
-			}
-			timePassed += ClusterReadInterval
-			fmt.Printf("\033[u\033[K%ds has elapsed while waiting for the cluster to reach RUNNING.\n", timePassed)
-			time.Sleep(ClusterReadInterval * time.Second)
+func waitClusterStatus(clusterId, hostname, username, apiKey, expectedStatus string, delayAfter time.Duration) error {
```
michael-carter-instaclustr, line 43: What is all this for? Isn't the capacity to wait already built into the provider?

acc_test/resource_kafka_test.go
h3nd24: OK, just realized there's a new function to suppress the diff now

acc_test/resource_kafka_test.go
h3nd24: not yet, hence all the discussions before about where this is gonna be put. I think along with this deployment, I will be adding the values to the CircleCI env

acc_test/test_utils.go
h3nd24: Historical reason for this one is this one predates the provider waiting stuff. The reason I used it now seems to be based on wrong assumption, I thought if I create 2 KC clusters at the same time and wait for running, it will create the cluster sequentially.

acc_test/resource_kafka_test.go
michael-carter-instaclustr: ok, don't forget to record themeless, because once they're in circleCI you won't be able to see them again

docs/resources/cluster.md
```diff
@@ -91,9 +91,9 @@ Property | Description | For Bundles | Default
...
 `vpc_type`|Available options: `KAFKA_VPC`, `VPC_PEERED`, `SEPARATE_VPC`. Only required if targeting an Instaclustr managed cluster.|Kafka Connect|`SEPARATE_VPC`
-`aws_access_key`, `aws_secret_key`, `s3_bucket_name`|Access information for the S3 bucket where you will store your custom connectors. (if using AWS)|Kafka Connect
-`azure_storage_account_name`, `azure_storage_account_key`, `azure_storage_container_name`|Access information for the Azure Storage container where you will store your custom connectors.|Kafka Connect
-`ssl_enabled_protocols`, `ssl_truststore_password`, `ssl_protocol`, `security_protocol`, `sasl_mechanism`, `sasl_jaas_config`, `bootstrap_servers`|Connection information for your Kafka Cluster. These options are analogous to the similarly named options that you would place in your Kafka Connect worker.properties file. Only required if connecting to a Non-Instaclustr managed Kafka Cluster|Kafka Connect
+`aws_access_key` (deprecated), `aws_secret_key` (deprecated), `s3_bucket_name`|Access information for the S3 bucket where you will store your custom connectors. (if using AWS)|Kafka Connect
```
michael-carter-instaclustr, line 94: so if they're deprecated, what have they been replaced with, and is the replacement mentioned anywhere?

instaclustr/resource_cluster.go
```diff
@@ -592,6 +592,45 @@ func resourceCluster() *schema.Resource {
...
 				},
 			},
+			"kafka_connect_credential": {
+				Type:	  schema.TypeList,
+				Optional: true,
+				MaxItems: 1,
+				Elem:	  &schema.Resource {
+					Schema: map[string]*schema.Schema {
+						"aws_access_key": {
+							Type:     schema.TypeString,
```
michael-carter-instaclustr, line 602: No GCP?

docs/resources/cluster.md
h3nd24: The replacement is below, there's an additional block of `kafka_connect_credential` that takes care of this. I'll add `see kafka_connect_credential` to make it clearer

instaclustr/resource_cluster.go
h3nd24: previously I was trying to make it in time for the prod freeze, so it's just what's existing. I'll add the GCP support for KC along with this PR too

h3nd24: Actually, maybe it's better to be done in another ticket

h3nd24: https://instaclustr.atlassian.net/browse/INS-17136

acc_test/data/kafka_connect_create_cluster.tf
```diff
@@ -0,0 +1,119 @@
+# kafka connect acceptance test. Using the same base kafka cluster as kafka user testing to minimize cluster creation
+provider "instaclustr" {
+    username = "%s"
+    api_key = "%s"
+    api_hostname = "%s"
+}
+
+resource "instaclustr_cluster" "kafka_cluster" {
```
michael-carter-instaclustr, line 8: Is this used at all? Looks like maybe this tests reuse some other Kafka Cluster?

acc_test/resource_kafka_test.go
```diff
@@ -7,13 +7,16 @@ import (
...
 	"strconv"
+
 	"testing"
 
 	"github.com/hashicorp/terraform/helper/resource"
 	"github.com/hashicorp/terraform/terraform"
 	"github.com/instaclustr/terraform-provider-instaclustr/instaclustr"
 )
 
+// we put the test here so that the Kafka cluster can be reused amongst bunch of other Kafka related tests
```
michael-carter-instaclustr, line 18: Not sure what this comment means?
TestKafkaResource in in resource_kafaka_test.go. Seems like a reasonable place to me?

instaclustr/resource_cluster.go
```diff
@@ -592,6 +592,45 @@ func resourceCluster() *schema.Resource {
 					return
 				},
 			},
+			"kafka_connect_credential": {
+				Type:	  schema.TypeList,
```
michael-carter-instaclustr, line 596: Why a list with one element? As opposed to a set or something like that?

acc_test/resource_kafka_test.go
h3nd24: ah right, what I meant there was that the Kafka Connect testing is bunched here even though logically it should be part of the resource_cluster_test

instaclustr/resource_cluster.go
h3nd24: Arjun gave the idea of the list with one element and that worked easily. I tried the TypeSet before and couldn't make it to work yet, must be some syntax stuff and I couldn't seem to find examples easily. Would there be a strong preference one over the other?

acc_test/data/kafka_connect_create_cluster.tf
h3nd24: so this one is used for the Kafka Connect clusters. And you are right, this Kafka cluster is reused from the other tests (user, ACL, and topic)

instaclustr/resource_cluster.go
michael-carter-instaclustr: Not particularly, just curious

acc_test/resource_kafka_test.go
michael-carter-instaclustr: ok, should probably mention Kafka Connect specifically if that's what its about

acc_test/data/kafka_connect_create_cluster.tf
michael-carter-instaclustr: so why is it written here again as a unique resource? SHouldn't it exist somewhere else already

h3nd24: refactored the `kafka_cluster` out and now Kafka resource tests refer to the cluster that is read from `kafka_user_create_cluster.tf`

instaclustr/resource_cluster.go
```diff
@@ -1542,10 +1584,22 @@ func resourceClusterDelete(d *schema.ResourceData, meta interface{}) error {
...
+}
+
+func clusterDeleteRetry(client APIClientInterface, d resourceDataInterface, id string) *resource.RetryError {
+	fmt.Printf("waiting for the dependencies of cluster %s to be cleaned\n", id)
+	// try to delete until successful or the error is not about dependency
 	err := client.DeleteCluster(id)
-	if err != nil {
-		return fmt.Errorf("[Error] Error deleting cluster: %s", err)
+	if err != nil && strings.HasPrefix(err.Error(), "Status code: 409") {
+		return resource.RetryableError(fmt.Errorf("[Error] Dependencies still exist for cluster %s", id))
```
YuWang24, line 1598: Just a question here, how does this retry work?

h3nd24: Apparently based on the API code, it will return code 409 if there are still dependencies. Now the API client code will wrap the error with string `Status code: X`. This means, we can check if there are still dependencies if the string match.

YuWang24 (approved):

michael-carter-instaclustr (approved):

PR #145, opened 2021-12-14
Adding GCP VPC Peering Support to the Terraform Provider

Kartik1925:
Internal reference - INS-16576

https://instaclustr.atlassian.net/wiki/spaces/IDS/pages/762150913/Mitigation+of+Security+Risks+of+Circle+CI+Integration+of+Instaclustr+s+Terraform+Provider


matthewmillett-instaclustr:
Code duplication stats are high, but acceptable given that the schema between AWS and GCP peering requests are similar. An option to resolve this would be to create a subset schema and reference it within AWS and GCP implementations, but I think this option hurts readability.

docs/resources/vpc_peering.md
```diff
@@ -6,7 +6,7 @@ description: |-
 ---
 
 ### Resource: `instaclustr_vpc_peering`  
-A resource for managing VPC peering connections on Instaclustr Managed Platform. This is only avaliable for clusters hosted with the AWS provider.  
+A resource for managing VPC peering connections on Instaclustr Managed Platform. This is avaliable for clusters hosted with the AWS provider and the GCP provider.
```
matthewmillett-instaclustr, line 9: I'd reword this to be `This is available for clusters hosted with AWS and GCP.` - typo in word available and cleaning up provider names

docs/resources/vpc_peering.md
```diff
@@ -41,12 +41,25 @@ Property | Description | Default
 `peer_region`| The Region code for the accepter VPC, if the accepter VPC is located in a Region other than the Region in which you make the request. | Not Required
 `aws_vpc_connection_id`| The ID of the VPC peering connection. | Computed
 
-#### Example
+#### Example aws
```
matthewmillett-instaclustr, line 44: I'd reword this to `AWS Example`

docs/resources/vpc_peering.md
```diff
@@ -41,12 +41,25 @@ Property | Description | Default
...
     cluster_id = "${instaclustr_cluster.example.cluster_id}"
     peer_vpc_id = "vpc-123456"
     peer_account_id = "1234567890"
     peer_subnet = "10.0.0.0/20"
 }
-```
\ No newline at end of file
+```
+
+#### Example GCP
```
matthewmillett-instaclustr, line 54: Reword to `GCP Example`

docs/resources/vpc_peering.md
```diff
@@ -41,12 +41,25 @@ Property | Description | Default
...
+```
+resource "instaclustr_GCPvpc_peering" "gcp_example" {
+
```
matthewmillett-instaclustr, line 57: Should remove this newline

examples/main.tf
```diff
@@ -3,19 +3,19 @@
 terraform {
   required_providers {
     instaclustr = {
-      source = "instaclustr/instaclustr"
+      //source = "instaclustr/instaclustr"
```
matthewmillett-instaclustr, line 6: I'd suggest reverting your changes here, all users who download this will want to run it against production systems, not local testing

examples/main.tf
```diff
@@ -48,6 +48,32 @@ resource "instaclustr_cluster" "example" {
...
+    bundle = "APACHE_CASSANDRA"
+    version = "apache-cassandra-3.11.8.ic2"
+    options = {
+      auth_n_authz = true
+    }
+  }
+}
+
+
+
```
matthewmillett-instaclustr, line 76: Make sure to clean this file up and remove some of these extra blank lines

instaclustr/Common_vpc_peering.go
```diff
@@ -0,0 +1,71 @@
+package instaclustr
```
matthewmillett-instaclustr, line 1: Should rename this file to be `common_vpc_peering.go` to maintain file name consistency

instaclustr/Common_vpc_peering.go
```diff
@@ -0,0 +1,71 @@
...
+	if len(idParts) != 2 || idParts[0] == "" || idParts[1] == "" {
+		return nil, fmt.Errorf("Unexpected format of ID (%q), expected <CLUSTER-ID>&<VPC-PEERING-ID>", d.Id())
+	}
+
+	d.Set("cluster_id", idParts[0])
+	d.Set("vpc_peering_id", idParts[1])
+	return []*schema.ResourceData{d}, nil
+}
+
+func resourceVpcPeeringCreat(d *schema.ResourceData, meta interface{}) (string, error) {
```
matthewmillett-instaclustr, line 45: Function name should be `resourceVpcPeeringCreate`

instaclustr/resource_gcp_vpc_peering.go
```diff
@@ -0,0 +1,164 @@
...
+		Importer: &schema.ResourceImporter{
+			State: resourceVpcPeeringStateImport,
+		},
+
+		Schema: map[string]*schema.Schema{
+
+			"name": {
+				Type:     schema.TypeString,
+				Required: true,
+			},
```
matthewmillett-instaclustr, line 27: Should add a newline after this block to make this consistent

instaclustr/resource_gcp_vpc_peering.go
```diff
@@ -0,0 +1,164 @@
...
+	if err != nil {
+		return fmt.Errorf("[Error] Error creating GCP VPC peering request: %s", err)
+	}
+	var id string
+	id, err = client.CreateVpcPeering(cdcID, jsonStr)
+	if err != nil {
+		return fmt.Errorf("[Error] Error in creating the cluster: %s", err)
+	}
+	d.SetId(id)
+	d.Set("vpc_peering_id", id)
```
matthewmillett-instaclustr, line 96: Is `id` and `vpc_peering_id` equivalent?

matthewmillett-instaclustr (changes requested):
Overall looks decent, you’ve committed `examples/.DS_Store` and `examples/.terraform.lock.hcl` which you probably shouldn’t, you may want to look at resetting those files in your git repo

acc_test/data/valid.tf
```diff
@@ -32,3 +32,27 @@ resource "instaclustr_cluster" "valid" {
...
+  }
+
+  bundle {
+    bundle = "APACHE_CASSANDRA"
+    version = "apache-cassandra-3.11.8.ic2"
+    options = {
+      auth_n_authz = true
+    }
+  }
+}
```
Jack-Kilrain, line 58: **[Desired]** Missing an EOF here, basically just add a newline after this

acc_test/resource_cluster_test.go
```diff
@@ -44,6 +44,36 @@ func AccClusterResourceTestSteps(t *testing.T, testAccProviders map[string]terra
...
+
+	resource.Test(t, resource.TestCase{
+		Providers:    testAccProviders,
+		CheckDestroy: testCheckResourceDeleted("gcp_valid", hostname, username, apiKey),
+		Steps: []resource.TestStep{
+			{
+				Config: oriConfig,
+				Check: resource.ComposeTestCheckFunc(
+					testCheckResourceValid("gcp_valid"),
+					testCheckResourceCreated("gcp_validgcp_valid", hostname, username, apiKey),
```
Jack-Kilrain, line 63: **[Question]** The first parameter here seems off. Does that resource name work? Seems wrong.

instaclustr/provider.go
```diff
@@ -23,13 +23,14 @@ func Provider() *schema.Provider {
...
-			"instaclustr_firewall_rule":  resourceFirewallRule(),
-			"instaclustr_vpc_peering":    resourceVpcPeering(),
-			"instaclustr_kafka_user":     resourceKafkaUser(),
-			"instaclustr_kafka_topic":    resourceKafkaTopic(),
-			"instaclustr_kafka_acl":      resourceKafkaAcl(),
+			"instaclustr_cluster":         resourceCluster(),
+			"instaclustr_encryption_key":  resourceEncryptionKey(),
+			"instaclustr_firewall_rule":   resourceFirewallRule(),
+			"instaclustr_vpc_peering_aws": resourceVpcPeering(),
+			"instaclustr_vpc_peering_gcp": resourceGCPVpcPeering(),
```
Jack-Kilrain, line 30: **[Question, State Upgrade]** Renaming the resource mapping to delineate between GCP and AWS might cause issues for existing AWS VPCs. Has that been considered? It seems like something that can be fixed for customers with existing VPC declarations with `terraform mv instaclustr_vpc_peering.<NAME> instaclustr_vpc_peering_aws.<NAME>` but you'd need to confirm that.

instaclustr/resource_gcp_vpc_peering.go
```diff
@@ -0,0 +1,166 @@
...
+					Type: schema.TypeString,
+				},
+				Optional:      true,
+				ConflictsWith: []string{"peer_subnet"},
+			},
+		},
+	}
+}
+
+func GCPresourceVpcPeeringCreate(d *schema.ResourceData, meta interface{}) error {
```
Jack-Kilrain, line 73: **[Minor,Desired]** Good to keep the naming of functions consistent. If we are using PascalCase for public visible funcs then it should be `GcpResourceVpcPeeringCreate`, for private funcs it should be camelCase

instaclustr/Common_vpc_peering.go
```diff
@@ -0,0 +1,71 @@
...
+	if len(idParts) != 2 || idParts[0] == "" || idParts[1] == "" {
+		return nil, fmt.Errorf("Unexpected format of ID (%q), expected <CLUSTER-ID>&<VPC-PEERING-ID>", d.Id())
+	}
+
+	d.Set("cluster_id", idParts[0])
+	d.Set("vpc_peering_id", idParts[1])
+	return []*schema.ResourceData{d}, nil
+}
+
+func VpcPeeringCreate(d *schema.ResourceData, meta interface{}) (string, error) {
```
Jack-Kilrain, line 45: **[Question]** Why does this method have public exposure and the others don't?

instaclustr/resource_gcp_vpc_peering.go
```diff
@@ -0,0 +1,166 @@
...
+		return fmt.Errorf("[Error] Error creating GCP VPC peering request: %s", err)
+	}
+
+	var jsonStr []byte
+	jsonStr, err = json.Marshal(createData)
+	if err != nil {
+		return fmt.Errorf("[Error] Error creating GCP VPC peering request: %s", err)
+	}
+	var id string
+	id, err = client.CreateVpcPeering(cdcID, jsonStr)
```
Jack-Kilrain, line 92: **[Maintainability,Required]** It's really hard to see the difference between this func `CreateVpcPeering` and `VpcPeeringCreate` used above. What is the difference? Why does this return an error for cluster creation if it is VPC creation?

instaclustr/resource_gcp_vpc_peering.go
```diff
@@ -0,0 +1,166 @@
...
+	if err != nil {
+		return fmt.Errorf("[Error] Error in creating the cluster: %s", err)
+	}
+	d.SetId(id)
+	d.Set("vpc_peering_id", id)
+	d.Set("cdc_id", cdcID)
+	var vpcPeering *GCPVPCPeering
+	vpcPeering, err = client.GCPReadVpcPeering(cdcID, id)
+	if err != nil {
+		return fmt.Errorf("[Error] Error in reading GCP VPC peering connection: %s", err)
```
Jack-Kilrain, line 102: **[Question]** Is this reading a connection or is it reading the VPC configuration?

instaclustr/resource_gcp_vpc_peering.go
```diff
@@ -0,0 +1,166 @@
...
+	}
+	err = GCPresourceVpcPeeringReadHelper(d, vpcPeering)
+	if err != nil {
+		return fmt.Errorf("[Error] Error reading VPC peering connection: %s", err)
+	}
+	return nil
+
+}
+
+func GCPresourceVpcPeeringReadHelper(d *schema.ResourceData, vpcPeering *GCPVPCPeering) error {
```
Jack-Kilrain, line 133: **[Maintainability,Desired]** The naming of this func doesn't really tell me what it does. I can see from the body that it maps a GCPVPCPeering resource instance to a schema, but I shouldn't need to read into the body of it to know that. Might be better to name it sometime like `MapGCPVPCPeeringToResource`

instaclustr/structs.go
```diff
@@ -176,6 +183,19 @@ type VPCPeering struct {
 	PeerRegion         string        `json:"peerRegion"`
 }
 
+type GCPVPCPeering struct {
+	ID                 string `json:"id"`
+	Name               string `json:"name"`
+	GCPVpcConnectionID string `json:"peerProjectId"`
```
Jack-Kilrain, line 189: **[Question]** I'm guessing `peerProjectId` is GCP's naming convention, but do we use that internally in the API orr do we use `GCPVpcConnectionID`? Would be good to keep them consistent

instaclustr/resource_vpc_peering.go
```diff
@@ -112,16 +92,16 @@ func resourceVpcPeeringCreate(d *schema.ResourceData, meta interface{}) error {
...
 	if err != nil {
 		return fmt.Errorf("[Error] Error creating cluster: %s", err)
 	}
 	d.SetId(id)
 	d.Set("vpc_peering_id", id)
 	d.Set("cdc_id", cdcID)
-
-	vpcPeering, err := client.ReadVpcPeering(cdcID, id)
+	var vpcPeering *VPCPeering
+	vpcPeering, err = client.ReadVpcPeering(cdcID, id)
```
Jack-Kilrain, line 104: **[Question]** Why do we need to explicitly state that `vpcPeering` is a pointer to a VPCPeering instance? I like that it's more descriptive, just wondering why.

instaclustr/resource_vpc_peering.go
```diff
@@ -77,32 +75,14 @@ func resourceVpcPeering() *schema.Resource {
...
-			break
-		}
-		if timePassed > WaitForClusterTimeout {
-			return fmt.Errorf("[Error] Timed out waiting for cluster to have the status 'PROVISIONED' or 'RUNNING'. Current cluster status is '%s'", latestStatus)
-		}
-		time.Sleep(ClusterReadInterval * time.Second)
-		timePassed += ClusterReadInterval
+	cdcID, err := VpcPeeringCreate(d, meta)
+	if err != nil {
+		return fmt.Errorf("[Error] Error creating VPC peering request: %s", err)
```
Jack-Kilrain, line 82: **[Question]** This might just be me lacking understanding, but this error message doesn't seem to line up with the func call above, this seems like it failed to create the VPC itself?

Jack-Kilrain (changes requested):

acc_test/resource_cluster_test.go
Kartik1925: Sorry, that is a type I am correcting it right now.

instaclustr/provider.go
Kartik1925: Yup, thats right I will change the existing AWS resource name back to instaclustr_vpc_peering.go

instaclustr/Common_vpc_peering.go
Kartik1925: There are methods which were common for GCP and AWS peering hence in order to bring down duplication this new common file that had identical functions was created

instaclustr/resource_gcp_vpc_peering.go
Kartik1925: Here the connection is being read.

instaclustr/structs.go
Kartik1925: ok have changed that to `PeerProjectID` keeping it consistent with camel casing

instaclustr/resource_vpc_peering.go
Kartik1925: For some reason it was giving me invalid type error till I stated that explicitly

instaclustr/resource_vpc_peering.go
Kartik1925: The `VpcPeeringCreate ` func has common elements that are used by both GCP and AWS for creating a peering request and that func return cdcID which is when null gives this failed to create VPC error message

Jack-Kilrain: So this function creates a request object? Not an actual request? That doesn't line up with what the function is by it's name.

docs/resources/vpc_peering.md
```diff
@@ -6,7 +6,7 @@ description: |-
 ---
 
 ### Resource: `instaclustr_vpc_peering`  
-A resource for managing VPC peering connections on Instaclustr Managed Platform. This is only avaliable for clusters hosted with the AWS provider.  
+A resource for managing VPC peering connections on Instaclustr Managed Platform. This is avaliable for clusters hosted with the AWS and GCP .
```
matthewmillett-instaclustr, line 9: Replace `This is avaliable for clusters hosted with the AWS and GCP .` with `This is avaliable for clusters hosted with AWS and GCP.`

docs/resources/vpc_peering.md
```diff
@@ -41,12 +41,24 @@ Property | Description | Default
 `peer_region`| The Region code for the accepter VPC, if the accepter VPC is located in a Region other than the Region in which you make the request. | Not Required
 `aws_vpc_connection_id`| The ID of the VPC peering connection. | Computed
 
-#### Example
+#### AWS Example
 ```
-resource "instaclustr_vpc_peering" "example_vpc_peering" {
+resource "_instaclustr_vpc_peering" "example_vpc_peering" {
```
matthewmillett-instaclustr, line 46: Should this start with an underscore? Shouldn't it be `"instaclustr_vpc_peering"`?

instaclustr/resource_vpc_peering.go
Jack-Kilrain: If you have two different read funcs for AWS and GCP and you are re-using the variable then that's probably why. Pointers are unconvertible once type fixated.

examples/main.tf
```diff
@@ -6,16 +6,16 @@ terraform {
...
+      version = ">= 1.14.8"
     }
   }
 }
 
 provider "instaclustr" {
-  username = "<Your instaclustr username here>"
-  api_key = "<Your provisioning API key here>"
+  username = ""
+  api_key = ""
```
matthewmillett-instaclustr, line 16: You should revert these changes here, keep the `<Your instaclustr username here>` prefills

instaclustr/resource_gcp_vpc_peering.go
Jack-Kilrain: Ok, in that case, the method should be `GCPReadVpcPeeringConnection`, it's fairly ambiguous otherwise. Also out of curiosity, what details are returned?

instaclustr/Common_vpc_peering.go
Jack-Kilrain: Ok makes sense.

instaclustr/resource_vpc_peering.go
matthewmillett-instaclustr: I think you should reword these error messages really, having identical error messages in 3 different places makes debugging pretty difficult

matthewmillett-instaclustr (changes requested):

examples/main.tf
```diff
@@ -6,16 +6,16 @@ terraform {
       source = "instaclustr/instaclustr"
       //Change the source as per below to work with a local development copy on terraform version >=13
       //source = "terraform.instaclustr.com/instaclustr/instaclustr"
-      version = ">= 1.0.0"
+      version = ">= 1.14.8"
```
Jack-Kilrain, line 9: **[Required]** This doesn't line up with the existing content. It should be `1.0.0`.

Makefile
```diff
@@ -3,7 +3,8 @@ BIN_NAME=terraform-provider-instaclustr
 
 
 # for VERSION, don't add prefix "v", e.g., use "1.9.8" instead of "v1.9.8" as it could break circleCI stuff
-VERSION=1.14.10
+
+VERSION=1.14.11
```
Jack-Kilrain, line 7: **[Required]** This is a feature addition, so I would assume it would be a good time for a minor version bump to `1.15.0`.

examples/main.tf
Jack-Kilrain, line 16: **[Required]** These should have the template values in them from before

examples/main.tf
```diff
@@ -48,6 +48,29 @@ resource "instaclustr_cluster" "example" {
...
+    name = "GCP"
+  }
+  rack_allocation = {
+    number_of_racks = 3
+    nodes_per_rack = 1
+  }
+
+  bundle {
+    bundle = "APACHE_CASSANDRA"
+    version = "apache-cassandra-3.11.8.ic2"
```
Jack-Kilrain, line 68: **[Required]** We don't publicly use the legacy version format (part of a change I made for TF provider, it's not 100% but it's the majority). You'll want to change this to `3.11.8`

examples/main.tf
```diff
@@ -108,6 +131,15 @@ resource "instaclustr_vpc_peering" "example_vpc_peering" {
   peer_subnets = toset(["10.0.0.0/20", "10.0.32.0/20"])
 }
 
+resource "instaclustr_vpc_peering_gcp" "example_vpc_peering" {
+
+  name="name"
+  peer_vpc_network_name = "network name"
+  peer_project_id = "projectId"
+  peer_subnets = toset(["10.10.0.0/16", "10.11.0.0/16"])
+  cluster_id = "${instaclustr_cluster.gcp_example.id}"
```
Jack-Kilrain, line 140: **[Question]** Do you need to wait for the cluster to provision to create a VPC peering connection?

instaclustr/Common_vpc_peering.go
```diff
@@ -0,0 +1,71 @@
...
+	vpcPeeringID := d.Get("vpc_peering_id").(string)
+	log.Printf("[INFO] Deleting VPC peering connection %s.", vpcPeeringID)
+	err = client.DeleteVpcPeering(cdcID, vpcPeeringID)
+	if err != nil {
+		return fmt.Errorf("[Error] Error deleting VPC peering connection: %s", err)
+	}
+
+	d.SetId("")
+	d.Set("vpc_peering_id", "")
+	d.Set("cdc_id", "")
```
Jack-Kilrain, line 29: **[Question]** What about setting the connection ID, or project ID as GCP term it?

instaclustr/resource_gcp_vpc_peering.go
```diff
@@ -0,0 +1,166 @@
...
+
+}
+
+func GCPcreateVpcPeeringRequest(d *schema.ResourceData) (CreateGCPVPCPeeringRequest, error) {
+	result := CreateGCPVPCPeeringRequest{
+		Name:               d.Get("name").(string),
+		PeerVPCNetworkName: d.Get("peer_vpc_network_name").(string),
+		PeerProjectID:      d.Get("peer_project_id").(string),
+	}
+	if _, isSet := d.GetOk("peer_subnets"); isSet {
```
Jack-Kilrain, line 160: **[Question]** What about the single `peer_subnet` schema field? Does that not get checked?

Jack-Kilrain (changes requested):

instaclustr/Common_vpc_peering.go
Kartik1925: `peer_project_id` has been set in the `resource_gvp_vpc_peering.go` file as this was only GCP specific and not common

instaclustr/resource_vpc_peering.go
Kartik1925: Just gonna change the error message to `[Error] Error creating VPC peering request object `

examples/main.tf
Kartik1925: Yup but we can also straight away create a peering to an already provisioned cluster by having the cluster id inside `cluster_id`

instaclustr/resource_gcp_vpc_peering.go
Kartik1925: Yes

instaclustr/resource_gcp_vpc_peering.go
Kartik1925: The read vpc peering func for aws which reads the connection is named `ReadVpcPeering` and keeping that in mind to I renamed this func GCPReadVpcPeering

instaclustr/resource_gcp_vpc_peering.go
Kartik1925: While `VpcPeeringCreate` has all common data and is being called here the `CreateVpcPeering` is actually creating the peering request after being called from the `api_client.go` file. Thanks for pointing out the error the message here is wrong and should be something like `Failed to call common create functions`

instaclustr/resource_gcp_vpc_peering.go
Kartik1925: GCP dosen't support single `peer_subnet` onle `peer_subnets` parameter is present in the API

instaclustr/resource_gcp_vpc_peering.go
Kartik1925: Done

instaclustr/resource_gcp_vpc_peering.go
Jack-Kilrain: Alrighty, then probably best to remove that from the schema then.

examples/main.tf
Jack-Kilrain: Right, so shouldn't the example resources for peering have the `wait_for_state=RUNNING` field enabled if you have both a cluster and VPC in one resource and provisioned via the same plan?

instaclustr/resource_gcp_vpc_peering.go
Kartik1925: Done

matthewmillett-instaclustr: If you wanted to, you could probably get rid of this check for `if _, isSet := d.GetOk("peer_subnets"); isSet {`, as `peer_subnets` should always be set for GCP.

examples/main.tf
Kartik1925: Done

Jack-Kilrain (approved):
LGTM

matthewmillett-instaclustr (dismissed):

carlsandland (changes requested):
you should not be committing binary .DS_Store file

matthewmillett-instaclustr (dismissed):

Jack-Kilrain (approved):

Jack-Kilrain (approved):

matthewmillett-instaclustr (approved):

Jack-Kilrain (approved):

matthewmillett-instaclustr (approved):

PR #153, opened 2022-02-07
Adding a firewall update location

michael-carter-instaclustr:


instaclustr/api_client.go
```diff
@@ -168,6 +168,21 @@ func (c *APIClient) CreateFirewallRule(data []byte, clusterID string) error {
 	return nil
 }
 
+func (c *APIClient) UpdateFirewallRule(data []byte, clusterID string) error {
+	url := fmt.Sprintf("%s/provisioning/v1/%s/firewallRules/", c.apiServerHostname, clusterID)
+	resp, err := c.MakeRequest(url, "PUT", data)
+	if err != nil {
+		return err
+	}
+	bodyText, err := ioutil.ReadAll(resp.Body)
```
Mitchel-Stig-Instaclustr, line 177: slight optimisation - could we move this inside the below if statement so we dont do this on success if its not needed

instaclustr/resource_firewall_rule.go
```diff
@@ -78,15 +72,26 @@ func resourceFirewallRuleCreate(d *schema.ResourceData, meta interface{}) error
...
+
+	jsonStr, err := makeFirewallRuleJson(d)
 	if err != nil {
 		return fmt.Errorf("[Error] Error creating firewall rule: %s", err)
 	}
 
 	err = client.CreateFirewallRule(jsonStr, d.Get("cluster_id").(string))
 	if err != nil {
-		return fmt.Errorf("[Error] Error creating firewall fule: %s", err)
+		return fmt.Errorf("[Error] Error creating firewall rule: %s", err)
```
Mitchel-Stig-Instaclustr, line 94: Ha!

Mitchel-Stig-Instaclustr (changes requested):

instaclustr/api_client.go
michael-carter-instaclustr: Not as it is, 'cos the body Text is put in the  error message

Jack-Kilrain: I think Mitch means to put the `ioutil.ReadAll(...)` in the if check below. Also, I think we should also handle checking the `err` var incase the body is unreadible:
```go
if resp.StatusCode != 202 {
	bodyText, err := ioutil.ReadAll(resp.Body)
	if err != nil {
		bodyText = "<unparsable body>"
	}
	return errors.New(fmt.Sprintf("Status code: %d, message: %s", resp.StatusCode, bodyText))
}
```

arjunrajinstaclustr (approved):

Mitchel-Stig-Instaclustr (approved):

arjunrajinstaclustr (approved):

Jack-Kilrain (approved):

PR #158, opened 2022-03-16
Add support for provisioning PgBouncer with PostgreSQL

minhnguyen-ic:
Internal reference - INS-18165.
CircleCI WILL fail until provisioning API changes are implemented. Do not run until then. 

TODO:

- Update error messages on acc_test 
- Update documentation link


tharinduamila-insta (changes requested):
Will approve after adding in the documentation link for pool_mode

instaclustr/structs.go
```diff
@@ -112,28 +113,29 @@ type DataCentreCreateRequest struct {
 }
 
 type AddonBundles struct {
-	Bundle  string `json:"bundle"`
-	Version string `json:"version"`
+	Bundle  string        `json:"bundle" mapstructure:"bundle"`
+	Version string        `json:"version" mapstructure:"version"`
+	Options BundleOptions `json:"options,omitempty" mapstructure:"options"`
```
arjunrajinstaclustr, line 118: Would be better to have an AddonBundleOptions struct, unless there's a reason we can't easily switch to that?

minhnguyen-ic: Nah, can easily switch to that.

minhnguyen-ic: Actually can't easily switch to that. When read, we have a `addOnBundles` but when creating they are all `bundle` and share the same type. So separating them will be a bit clunky since there are 2 options struct to maintain. In this case, everything in AddonBundleOptions will need to be included in BundleOptions as well.

docs/resources/cluster.md
```diff
@@ -11,111 +10,113 @@ A resource for managing clusters on Instaclustr Managed Platform. A cluster cont
...
+| `pool_mode`                                                                                                                                                                                           | The connection pooling mode to be used with the PgBouncer add-on. The available options are `TRANSACTION` and `SESSION`. Refer to our [documentation](https://www.instaclustr.com/support/documentation/postgresql-add-ons/using-pgbouncer/) for more detail.                     | PgBouncer                                                                        | Required                                 |
```
minhnguyen-ic, line 112: @tharinduamila-insta Document link updated

arjunrajinstaclustr (dismissed):

tharinduamila-insta (approved):

yallen-ic (approved):

arjunrajinstaclustr (approved):

Quintero-daniel (approved):

PR #155, opened 2022-03-08
INS-17949 Add synchronous_mode_strict option for PostgreSQL clusters

alwyndavis:

PR #160, opened 2022-03-23
INS 18272 Add Cadence Support

RussDavidson-InstaClustr:
Internal Reference: INS-18272


tharinduamila-insta (approved):
PostgreSQL specific options doco update looks good to me. Please make sure if you are merging after #158 or #155 to fixup the related doco as well if they are listed as optional

instaclustr/resource_cluster.go
```diff
@@ -1308,6 +1348,7 @@ func resourceClusterRead(d *schema.ResourceData, meta interface{}) error {
 	d.SetId(cluster.ID)
 	d.Set("cluster_id", cluster.ID)
 	d.Set("cluster_name", cluster.ClusterName)
+	d.Set("cdc_id", cluster.CdcId)
```
arjunrajinstaclustr, line 1351: Not sure if "CDC" is really a public friendly term. Why not `data_centre_id` ? 
Same goes for `target_*_cdc_id`s

Should also document the read only data centre ID field under docs/resources/cluster.md

docs/resources/cluster.md
```diff
@@ -102,7 +103,14 @@ Property | Description | For Bundles | Default
...
-`replication_mode` | The default replication behaviour for PostgreSQL cluster. See [Replication Mode](https://www.instaclustr.com/support/documentation/postgresql/options/replication-mode/) documentation for more details. Allowed values are `SYNCHRONOUS` and `ASYNCHRONOUS`. |Postgresql| `SYNCHRONOUS`
+`replication_mode` | The default replication behaviour for PostgreSQL cluster. See [Replication Mode](https://www.instaclustr.com/support/documentation/postgresql/options/replication-mode/) documentation for more details. Allowed values are `SYNCHRONOUS` and `ASYNCHRONOUS`. |Postgresql|Required
+`advanced_visibility`|Accepts true/false. Enables Advanced Visibility features for Cadence.|Cadence|false
+`target_cassandra_cdc_id`|GUID of the Instaclustr managed Cassandra Data Centre Id you wish to connect to. Must be in the same Instaclustr account.|Cadence|Required
+`target_cassandra_vpc_type`|Available options: `SEPARATE_VPC`, `CASSANDRA_CLUSTER_VPC`, `VPC_PEERING`|Cadence|Required
+`target_opensearch_cdc_id`|GUID of the Instaclustr managed Opensearch Cluster Id you wish to connect to. Must be in the same Instaclustr account.|Cadence|Required if Advanced Visibility enabled.
```
arjunrajinstaclustr, line 110: Is this a cluster ID? Or Data Centre ID? because the description and property names don't match.
Same for `target_kafka_cdc_id`

arjunrajinstaclustr (changes requested):

instaclustr/resource_cluster.go
arjunrajinstaclustr: Since there's a mix of multi-DC and single DC schemas in this resource, this should probably even be `default_data_centre_id` to make it clear.

RussDavidson-InstaClustr: Hey Arjun,

Re: "Should also document the read only data centre ID field under docs/resources/cluster.md"

Are you thinking that this should be in the "Schema" section on that page? Those appear to be input fields, and not read only documented there though, as there are other fields that are read only/computed and not documented there though, such as 'cluster_id'.

Happy to add it either way, just wanted to check.

Thanks,
Russ

arjunrajinstaclustr: Yeah we should document read-only properties too. You could just mention that they are Read-only in the description and default columns. Other read-only inputs definitely should have also been documented :)

PR #172, opened 2022-06-02
Opensearch132

jacobhuanginstaclustr:
Updating to OpenSearch 1.3.2 for terraform providers. 

Internal reference: INS-19458


yiwei-wu-instaclustr:
1. the tests fail because 1.3.2 branch is not in app yet
2. double check the terraform version before deployment, I saw the other open PR has bumped the version as well

yiwei-wu-instaclustr:
We probably need the 2.0.0 in https://github.com/instaclustr/terraform-provider-instaclustr/blob/5fc354673c8fe99d3c38807a902a578a68334c8d/docs/resources/cluster.md as well.

yiwei-wu-instaclustr (approved):

acc_test/import_resource_test.go
```diff
@@ -131,39 +131,39 @@ func TestAccKafkaCluster_importBasic(t *testing.T) {
...
-			{
-				Config:            oriConfig,
-				ResourceName:      "instaclustr_encryption_key.valid",
-				ImportState:       true,
-				ImportStateVerify: true,
-			},
-		},
-	})
-}
+//func TestAccEncryptionKey_importBasic(t *testing.T) {
```
arjunrajnetapp, line 134: Why comment these out?

acc_test/data/valid_with_vpc_peering.tf
```diff
@@ -71,7 +71,7 @@ resource "instaclustr_cluster" "gcp_example" {
 
   bundle {
     bundle = "APACHE_CASSANDRA"
-    version = "3.11.8"
```
arjunrajnetapp, line 74: These older versions were applied as bundle overrides to the test account, so if we are updating the version, be sure to apply overrides for the new version too so we aren't constantly updating the bundle versions as we roll out new versions

jacobhuanginstaclustr:
>

jacobhuanginstaclustr (commented):
dismissed due to accidental click on review.

acc_test/import_resource_test.go
jacobhuanginstaclustr: I think this is one of those tests that has been failing constantly on CircleCI due to “after applying the plan there was no change” something like that.

acc_test/data/valid_with_vpc_peering.tf
jacobhuanginstaclustr: I changed this because I saw someone changed the version of a “find and replace” command in some tests probably without realising what they changed? So, basically these versions changed in the data folder are for matching the “find” part of the “find and replace” command otherwise whatever people wants to replace into those templates in the data folder would not be affected because the specified version in that command is never found. That’s my understanding though which could be wrong. Please lemme know if that’s the case and I’ll push a patch to them.

arjunrajnetapp: The "find and replace" feature, which comes from `AccClusterResourceTestSteps` which isn't used by most of these changed files from the looks of it, it is used in `valid.tf` though so that certainly needs to be updated - thanks for that. 

For these other files, it's fine to use the newer version as long as the tests are passing, and we have a bundle version override for it on the test account so that these tests don't need to be updated when we release a new bundle version.

acc_test/import_resource_test.go
arjunrajnetapp: Do you have a link to the failed Circle CI job to confirm if we've got a bug?

Mitchel-Stig-Instaclustr, line 134: Maybe @arjunrajnetapp can confirm, but is the process to just comment out the tests? shouldnt we be deleting them?  commented code always gives me shivers

acc_test/import_resource_test.go
```diff
@@ -257,8 +257,8 @@ func TestKafkaUserResource_importBasic(t *testing.T) {
 			},
 			{
 				PreConfig: func() {
-					fmt.Println("Sleep for 6 minutes to wait for Kafka cluster to be ready for resize.")
-					time.Sleep(6 * time.Minute)
+					fmt.Println("Sleep for 10 minutes to wait for Kafka cluster to be ready for resize.")
```
Mitchel-Stig-Instaclustr, line 260: We have to be careful of this, the tests have a hard stop at the 2 hour mark (if my memory serves correctly) at which point they fail.  Not telling you to remove, just letting you know.

README.md
```diff
@@ -285,7 +285,7 @@ Configuration documentation can be found at the [Instaclustr Terraform Registry]
 | KAFKA_REST_PROXY         | 5.0.0                                 | KAFKA                                                                              |
 | KAFKA_SCHEMA_REGISTRY    | 5.0.0, 5.0.4                          | KAFKA                                                                              |
 | KARAPACE_SCHEMA_REGISTRY | 2.1.2                                 | KAFKA <br/> **Not compatible with:** <br/> KAFKA_REST_PROXY, KAFKA_SCHEMA_REGISTRY |
-| OPENSEARCH               | 1.2.4                                 |                                                                                    |
+| OPENSEARCH               | 1.3.2, 2.0.0                          |                                                                                    |
 | ELASTICSEARCH (For Legacy Support Only)           | 1.13.3                                |                                                                                    |
```
Mitchel-Stig-Instaclustr, line 289: could probably fix up this line to be in-line with the rest

acc_test/import_resource_test.go
arjunrajnetapp: We'll be getting rid of all of these tests come v2, but in my opinion I'd prefer to comment these out than outright delete them in the off chance that we want to turn them back on between now and releasing v2.

acc_test/import_resource_test.go
arjunrajnetapp: 1 hour mark actually :) 
Since we've removed a whole bunch of tests that revolved around dependencies, we should have freed up a lot of time in our tests so we might be able to afford adding 4 minutes here. Ideally though we would be doing a polled check with a timeout limit rather than a hard wait like this. On further thought I think I'd rather comment this test out than have to add to the wait.

acc_test/import_resource_test.go
jacobhuanginstaclustr: Think this one test specifically is commented out by mistake. But this is a link to one of the jobs with failing tests that I'm referring to here: https://app.circleci.com/pipelines/github/instaclustr/terraform-provider-instaclustr/896/workflows/7b5445c1-af7e-4290-b930-3a70a919a1cb/jobs/1292

acc_test/data/valid_with_vpc_peering.tf
jacobhuanginstaclustr: Oh great that's good to know. I do realise that i have commented out a few other tests that were using some templates in the "data" folder and there's no point making those changes anymore. But I also remember before commenting them out, they were failing on the replace command so that's why i made those changes in the first place.

README.md
jacobhuanginstaclustr: I'm not exactly sure what you mean by in-line here Mitch?

Mitchel-Stig-Instaclustr (approved):

yiwei-wu-instaclustr (dismissed):

Mitchel-Stig-Instaclustr (approved):

Mitchel-Stig-Instaclustr (approved):

iclonghoang (approved):

PR #174, opened 2022-06-16
Terraform support to Update PrivateLink Endpoint Service Principal ARNs

sujayinstaclustr:
1. Added API client method to update endpoint service Principal ARNS
2. Changes to Terraform Cluster Update to call API Client method to update the endpoint service principal ARNs


arjunrajinstaclustr (approved):

instaclustr/resource_cluster.go
```diff
@@ -352,7 +352,7 @@ func resourceCluster() *schema.Resource {
 								Type: schema.TypeString,
 							},
 							Optional: true,
-							ForceNew: true,
+							ForceNew: false,
```
arjunrajinstaclustr, line 355: We might want to update the registry documentation to mention that this field is updatable, like we do for node_size, kafka_rest_proxy_user_password and kafka_schema_registry_user_password

sujayinstaclustr: I think I created the branch on the instaclustr repo instead of forked repo. I will close this PR and create a new one from my fork

sujayinstaclustr: Updated the `docs/resources/cluster.md` documentation file

instaclustr/api_client.go
```diff
@@ -390,3 +390,23 @@ func (c *APIClient) UpdateBundleUser(clusterID string, bundle string, data []byt
 	}
 	return nil
 }
+
+func (c *APIClient) UpdateEndpointServicePrincipals(clusterDataCentreID string, data []byte) error {
```
YuWang24, line 394: Can we have a unit test to cover this func? But I saw Arjun have already approved it, so I guess it's up to you now. Not very necessary.

docs/resources/cluster.md
```diff
@@ -53,7 +53,7 @@ A resource for managing clusters on Instaclustr Managed Platform. A cluster cont
 
 | Property             | Description                                                    | Default  |
 |----------------------|----------------------------------------------------------------|----------|
-| `iam_principal_arns` | List of IAM Principal ARNs to add to the endpoint service      | Required |
+| `iam_principal_arns` | List of IAM Principal ARNs to add to the endpoint service. This field is updatable      | Required |
```
YuWang24, line 56: I think it's not **Required** anymore after Vu's api changes? But if someone only has an empty `privateLink{}` block, Terraform won't read this block in, and therefore that CDC won't set PrivateLink enabled. So it's more like **Required, but can be an empty list**.

YuWang24 (changes requested):

instaclustr/api_client.go
sujayinstaclustr: Makes sense I will include a unit test.

YuWang24 (dismissed):

YuWang24 (dismissed):

logan-instaclustr (dismissed):

arjunrajinstaclustr (dismissed):

arjunrajinstaclustr (approved):

weixiong-ins (approved):

arjunrajinstaclustr (approved):

YuWang24 (approved):

arjunrajnetapp (approved):

YuWang24 (approved):

PR #179, opened 2022-07-07
INS-19692 - kafka version 3.1.1 ga

vutdv-ic:
add reference to kafka 3.1.1 and kafka connect 3.1.1
update example to latest version

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -5,7 +5,7 @@ BIN_NAME=terraform-provider-instaclustr
 # for VERSION, don't add prefix "v", e.g., use "1.9.8" instead of "v1.9.8" as it could break circleCI stuff
 
 
-VERSION=1.25.0
+VERSION=1.26.0
 
 INSTALL_FOLDER=$(HOME)/.terraform.d/plugins/terraform.instaclustr.com/instaclustr/instaclustr/$(VERSION)/darwin_amd64
 
--- a/Makefile
+++ b/Makefile
@@ -5,7 +5,7 @@ BIN_NAME=terraform-provider-instaclustr
 # for VERSION, don't add prefix "v", e.g., use "1.9.8" instead of "v1.9.8" as it could break circleCI stuff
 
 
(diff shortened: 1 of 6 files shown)


Mitchel-Stig-Instaclustr (dismissed):

examples/main.tf
```diff
@@ -542,7 +542,7 @@ resource "instaclustr_cluster" "example-cadenceav-kafka" {
 
   bundle {
     bundle = "KAFKA"
-    version = "3.0.0"
+    version = "3.1.1"
```
YuWang24, line 545: There's no need to change all the versions to 3.1.1. Especially for Cadence, if we want to change this, we'd better to test it whether this version can work well with Cadence.

sujayinstaclustr: @vutdv-ic  @YuWang24  is right Cadence has a pinned version of Kafka (3.0.0) which is a hard dependency on the Kafka version. Don't change this unless Cadence team wants to change it.

vutdv-ic: right, updated

Makefile
```diff
@@ -5,7 +5,7 @@ BIN_NAME=terraform-provider-instaclustr
 # for VERSION, don't add prefix "v", e.g., use "1.9.8" instead of "v1.9.8" as it could break circleCI stuff
 
 
-VERSION=1.25.0
+VERSION=1.26.0
```
arjunrajnetapp, line 8: Should be a patch version bump as per semver rules as this is pretty much just documentation changes and test fixes, and we aren't adding any new features.

arjunrajnetapp (changes requested):

arjunrajnetapp (dismissed):

sujayinstaclustr (dismissed):

YuWang24 (dismissed):

YuWang24 (approved):

arjunrajnetapp (approved):

PR #191, opened 2022-08-30
Provider v2 automated documentation + simplifications to build

arjunrajnetapp:
Internal reference: INS-17900


scripts/instaclustr-terraform-registry-documentation-update.sh
```diff
@@ -0,0 +1,25 @@
+#!/usr/bin/env bash
+
+set -e
+
+INSTACLUSTR_API_URL="$IC_API_URL"
+
+if [ -z "$INSTACLUSTR_API_URL" ]
+then
+  INSTACLUSTR_API_URL="https://api.instaclustr.com"
```
michael-carter-instaclustr, line 9: Isn't localhost going to be the more used default?

main.go
```diff
@@ -8,23 +8,17 @@ import (
...
-//go:generate go run github.com/hashicorp/terraform-plugin-docs/cmd/tfplugindocs
-
-var ProviderHost = "https://api.instaclustr.com"
-
 const ProviderName = "instaclustr"
 
 func main() {
 
-	swaggerHostOverride := os.Getenv("IC_TF_PROVIDER_SWAGGER_HOST_OVERRIDE") //needed to target locally run API to generate docs for testing/preview
+	providerHost := os.Getenv("IC_API_URL") //needed to target locally run API to generate docs for testing/preview
```
michael-carter-instaclustr, line 15: won't this always overwrite what was passed in from the makefile?

arjunrajnetapp: No it won't, the URL isn't baked in on the built binary. The built binary will only override when it detects the `IC_API_URL` environment variable.

scripts/instaclustr-terraform-registry-documentation-update.sh
arjunrajnetapp: Probably, yeah. But for production safety, the default is the production URL. In any case though, there's a simple make command to use to generate the docs for local testing - `make local-gen-docs`

Mitchel-Stig-Instaclustr (approved):

PR #200, opened 2022-10-02
INS-20846-provision-m-tls-enabled-kafka-cluster-via-api-2

vutdv-ic:
update with make local-gen-docs


debaditb (commented):
There seems to be a lot of formatting changes, Can we may be undo those?

vutdv-ic:
> There seems to be a lot of formatting changes, Can we may be undo those?

yeah, this one is done by using

make local-gen-docs

so will wait for Customer Platform on this, since this also new to me

PR #215, opened 2022-10-31
INS-21105 apiv2 zookeeper bundle changes

jacobhuanginstaclustr:
Generating zookeeper docs for API v2


vutdv-ic (commented):
we also need to include version bump in Makefile, make sure to based on latest provider-v2 branch

vutdv-ic (approved):

docs/data-sources/zookeeper_cluster_v2_instance.md
```diff
@@ -0,0 +1,169 @@
...
+<br>List of tags to apply to the Data Centre. Tags are metadata labels which  allow you to identify, categorize and filter clusters. This can be useful for grouping together clusters into applications, environments, or any category that you require.<br><br>
+*___name___*<br>
+<ins>Type</ins>: string, read-only<br>
+<br>A logical name for the data centre within a cluster. These names must be unique in the cluster.<br><br>
+*___nodes___*<br>
+<ins>Type</ins>: repeatable nested block, read-only, see [nodes](#nested--nodes) for nested schema<br>
+<br>
+*___node_size___*<br>
+<ins>Type</ins>: string, read-only<br>
+<br>Size of the nodes provisioned in the Data Centre. --AVAILABLE_NODE_SIZES_MARKER_V2_ZOOKEEPER--<br><br>
```
debaditya-netapp, line 91: These one should be `AVAILABLE_NODE_SIZES_MARKER_V2_APACHE_ZOOKEEPER`, the `BundleType` for zookeeper. Then it will be replaced by the available node sizes for that bundle while the docs are being generated.

debaditya-netapp:
In general I would suggest to use the BundleType name for Zookeeper, which is `APACHE_ZOOKEEPER` instead of `ZOOKEEPER`

docs/resources/kafka_topic_v2.md
```diff
@@ -0,0 +1,62 @@
+---
```
debaditya-netapp, line 1: I think we don't need to add this file as part of this PR, as its unrelated.

docs/data-sources/zookeeper_cluster_v2_instance.md
jacobhuanginstaclustr: Seems this is generated from the source yaml file and since i didn't see this comment and that file has been merged I couldn't change it now.

jacobhuanginstaclustr:
Closing due to already been merged in other PRs.

PR #220, opened 2022-11-17
Write only property fixes

arjunrajnetapp:
Updated openapi plugin dependency containing write only property fixes


Makefile
```diff
@@ -1,6 +1,6 @@
 BIN_NAME=terraform-provider-instaclustr
 
-VERSION=2.0.9
+VERSION=2.0.10
```
niluka-insta, line 3: I am making this 2.0.10 in https://github.com/instaclustr/terraform-provider-instaclustr/pull/219/files to test the deployment pipeline. 
Do you want to change this or should we be taking this together?

niluka-insta (approved):

PR #236, opened 2023-01-10
INS 21572 role based access to KC via APIv2

jacobhuanginstaclustr:


debaditya-netapp (approved):

Makefile
```diff
@@ -1,6 +1,6 @@
 BIN_NAME=terraform-provider-instaclustr
 
-VERSION=2.0.15
+VERSION=1.29.1
```
arjunrajnetapp, line 3: Should be v2.0.16

arjunrajnetapp (changes requested):

WeiXinFam-Insta (commented):
Where's the info on if no access key/secret key nor iam role provided, will be the attach permission policy method by default?

jacobhuanginstaclustr:
> 

It is included in the resource documentation.

Makefile
jacobhuanginstaclustr: Fixed it :)

arjunrajnetapp (approved):

PR #242, opened 2023-01-22
Internal reference removal

arjunrajnetapp:
Ticket reference: INS-21362


docs-base/index.md
```diff
@@ -21,7 +21,7 @@ This provider is in __General Availability__ with support for the following offe
 | Apache ZooKeeper | Complete    |
 | OpenSearch | In Progress    |
 | Redis | In Progress    |
-| PostgreSQL | In Progress    |
+| PostgreSQL | Complete    |
```
niluka-insta, line 24: Is this done opportunistically? has the postgres team forgotten to update this?

arjunrajnetapp: Yes that's right - the team updated the generated output but not the source of the content which is from this folder.

niluka-insta (approved):

PR #263, opened 2023-03-03
Ins 21379 moved terraform docs for OpenSearch back

vietic:


vietic:
close due to out of date

PR #332, opened 2023-09-11
Ins 24675

jackwalkernetapp:


iclonghoang (commented):
looks like we remove documentation of deprecated resources, e.g., userV2
maybe this is an issue in our gen-local-doc script

anyway, not much we can do immediately with this

iclonghoang (approved):

debaditya-netapp (approved):

PR #352, opened 2023-11-17
INS-26409

ruhirInstaclustr:


docs/data-sources/cassandra_cluster_v2_instance.md
anamarim-instaclustr, line 1: AFAIK, we should include just the Kafka and Kafka Connect docs in our TF PRs.

